PR #59, opened 2018-03-20
Feature/add tests for immutable cluster parameters

kvahed:
Added test for immutable fields.
Rest is to follow in later, derived PR.


tests/immutable_test.go
```diff
@@ -0,0 +1,73 @@
+package tests
```
ewoutp: Add file header

PR #65, opened 2018-03-22
Integrate with scaling web-UI

ewoutp:


pkg/deployment/cluster_informer.go
```diff
@@ -0,0 +1,104 @@
...
+		// #DBServers has changed
+		dbserversChanged = true
+	}
+	if !coordinatorsChanged && !dbserversChanged {
+		// Nothing has changed
+		return nil
+	}
+	// Let's update the spec
+	current, err := d.deps.DatabaseCRCli.DatabaseV1alpha().ArangoDeployments(d.apiObject.Namespace).Get(d.apiObject.Name, metav1.GetOptions{})
+	if err != nil {
```
kvahed: Maybe a misunderstanding and also premature optimisation. Couldn't you release the mutex already around here?
I was assuming that the update further down would be guarding itself?

ewoutp: That would be ok, but makes the rest of the code more complicated because you cannot unlock in defer AND here.

PR #64, opened 2018-03-22
add test for different deployments

ObiWahn:

PR #69, opened 2018-03-23
Test/different deployments resilient

ObiWahn:
add test for resilient single


ewoutp:
Make sure to merge in master in here. See #64 for why

ewoutp:
Add a comment where help is needed

ewoutp:
Code LGTM, since last jenkins test failed, I've triggered another one with LONG enabled.
If that succeeds, we can merge.

PR #72, opened 2018-03-26
Feature/test individual pod deletion

ewoutp:


tests/resilience_test.go
```diff
@@ -0,0 +1,239 @@
...
+	api "github.com/arangodb/kube-arangodb/pkg/apis/deployment/v1alpha"
+	"github.com/arangodb/kube-arangodb/pkg/client"
+	"github.com/arangodb/kube-arangodb/pkg/util/retry"
+)
+
+// TestResiliencePod
+// Tests handling of individual pod deletions
+func TestResiliencePod(t *testing.T) {
+	longOrSkip(t)
+	c := client.MustNewInCluster()
```
neunhoef: This panics in case of failure, is this the right behaviour? Should we not rather hand in the *testing.T and call Fatalf in that case?

PR #73, opened 2018-03-26
Image upgrade downgrade test

ewoutp:


ewoutp:
This test is now obsolete and part of the existing upgrading tests.

PR #76, opened 2018-03-26
add upgrade tests

ObiWahn:


tests/test_util.go
```diff
@@ -297,3 +297,69 @@ func removeSecret(cli kubernetes.Interface, secretName, ns string) error {
 	}
 	return nil
 }
+
+func arangoDeploymentHealthy(t *testing.T, deployment *api.ArangoDeployment, k8sClient kubernetes.Interface) {
+	// Create a database client
+	ctx := context.Background()
+	DBClient := mustNewArangodDatabaseClient(ctx, k8sClient, deployment, t)
```
ewoutp: Prefer to keep the dbclient as an argument passed to this function.

tests/test_util.go
```diff
@@ -297,3 +297,69 @@ func removeSecret(cli kubernetes.Interface, secretName, ns string) error {
 	}
 	return nil
 }
+
+func arangoDeploymentHealthy(t *testing.T, deployment *api.ArangoDeployment, k8sClient kubernetes.Interface) {
```
ewoutp: Add comment.

ewoutp: Name is now clean to me.
Suggest to rename to `waitUntilArangoDeploymentHealthy`

ewoutp: To be in line with other `waitUntil...` functions, this function must return an `error` in case something is wrong.
You can (read should) then remove the `t *testing.T` argument.

tests/upgrade_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+	"github.com/dchest/uniuri"
+
+	driver "github.com/arangodb/go-driver"
+	api "github.com/arangodb/kube-arangodb/pkg/apis/deployment/v1alpha"
+	kubeArangoClient "github.com/arangodb/kube-arangodb/pkg/client"
+	"github.com/arangodb/kube-arangodb/pkg/util"
+)
+
+// TODO - environments (provided from outside)
```
ewoutp: I don't understand. What is there TODO?

Perhaps you can replace this by a comment, saying that the kubernetes deployment environments are controlled outside the tests and are therefor not a variable in these tests.

tests/upgrade_test.go
```diff
@@ -0,0 +1,122 @@
...
+		t.Fatalf("Create deployment failed: %v", err)
+	}
+
+	// Wait for deployment to be ready
+	deployment, err = waitUntilDeployment(deploymentClient, deploymentTemplate.GetName(), k8sNameSpace, deploymentHasState(api.DeploymentStateRunning))
+	if err != nil {
+		t.Fatalf("Deployment not running in time: %v", err)
+	}
+
+	arangoDeploymentHealthy(t, deployment, k8sClient)
```
ewoutp: Comment please

tests/upgrade_test.go
```diff
@@ -0,0 +1,122 @@
...
+			spec.Image = util.NewString("arangodb/arangodb:" + toVersion)
+		})
+	if err != nil {
+		t.Fatalf("Failed to upgrade the Image from version : " + fromVersion + " to version: " + toVersion)
+	}
+
+	arangoDeploymentHealthy(t, deployment, k8sClient)
+
+	ctx := context.Background()
+	DBClient := mustNewArangodDatabaseClient(ctx, k8sClient, deployment, t)
```
ewoutp: I think this test must be done on all servers.
The current implementation of the test can succeed even f not all servers are actually upgraded.

ewoutp: Also, you put this test in an `op := func() error {... ` with a `Retry`: "Wait until all versions are correct"
Reason: It could be that the change of image is saved, the cluster is still healthy and the version check test
starts before any upgrading has actually taken place.

tests/test_util.go
```diff
@@ -231,9 +236,30 @@ func waitUntilVersionUp(cli driver.Client, allowNoLeaderResponse ...bool) error
 		return maskAny(noLeaderErr)
 	}
 
+	if predicate != nil {
```
ewoutp: Move the predicate call inside the `op` function.
I expect the function to wait until the version response is valid, including any predicate I may pass in.

tests/test_util.go
```diff
@@ -231,9 +236,30 @@ func waitUntilVersionUp(cli driver.Client, allowNoLeaderResponse ...bool) error
...
+	return func(infoFromServer driver.VersionInfo) error {
+		if (info.Version.CompareTo(infoFromServer.Version)) != 0 {
+			return maskAny(fmt.Errorf("given version %v and version from server %v do not match", info.Version, infoFromServer.Version))
+		}
+		return nil
+	}
+}
+
+// creates predicate to be used in waitUntilVersionUp
+func createEqualVersionsPredicateFromString(version_string string) func(driver.VersionInfo) error {
```
ewoutp: Avoid underscores in names in go.
Use camelCasing

tests/test_util.go
```diff
@@ -231,9 +236,30 @@ func waitUntilVersionUp(cli driver.Client, allowNoLeaderResponse ...bool) error
...
 
+	if predicate != nil {
+		return predicate(vInfo)
+	}
+
 	return nil
 }
 
+// creates predicate to be used in waitUntilVersionUp
+func createEqualVersionsPredicate(info driver.VersionInfo) func(driver.VersionInfo) error {
```
ewoutp: I would expect this function to have a `version driver.Version` argument, since that is the only thing you're comparing.
Will automatically make `createEqualVersionsPredicateFromString` obsolete.

tests/test_util.go
```diff
@@ -297,3 +323,72 @@ func removeSecret(cli kubernetes.Interface, secretName, ns string) error {
 	}
 	return nil
 }
+
+func waitUntilArangoDeploymentHealthy(deployment *api.ArangoDeployment, DBClient driver.Client, k8sClient kubernetes.Interface, versionString ...string) error {
```
ewoutp: Add comment describing what the function does.

ewoutp: Replace `versionString ...string` with `requiredVersion driver.Version`.
If that is a non-empty string, create the predicate.

Reason: I think we may want to add other checks in the future and you can only have 1 optional argument.

tests/test_util.go
```diff
@@ -297,3 +323,72 @@ func removeSecret(cli kubernetes.Interface, secretName, ns string) error {
...
+	case api.DeploymentModeResilientSingle:
+		if err := waitUntilVersionUp(DBClient, checkVersionPredicate); err != nil {
+			return maskAny(fmt.Errorf("Single Server not running in time: %v", err))
+		}
+
+		members := deployment.Status.Members
+		singles := members.Single
+		agents := members.Agents
+
+		if len(singles) != 2 || len(agents) != 3 {
```
ewoutp: Don't like these constants. 
Check against a spec.

tests/test_util.go
```diff
@@ -297,3 +323,72 @@ func removeSecret(cli kubernetes.Interface, secretName, ns string) error {
...
+			if err := waitUntilVersionUp(dbclient, checkVersionPredicate, true); err == nil {
+				goodResults++
+			} else if driver.IsNoLeader(err) {
+				noLeaderResults++
+			} else {
+				return maskAny(fmt.Errorf("Version check failed for: %v", single.ID))
+			}
+		}
+
+		if goodResults < 1 || noLeaderResults > 1 {
```
ewoutp: Don't like the last `1` constant. Check against a spec.

tests/upgrade_test.go
```diff
@@ -0,0 +1,124 @@
...
+func upgradeSubTest(t *testing.T, mode api.DeploymentMode, engine api.StorageEngine, fromVersion, toVersion string) error {
+	// check environment
+	longOrSkip(t)
+
+	k8sNameSpace := getNamespace(t)
+	k8sClient := mustNewKubeClient(t)
+	deploymentClient := kubeArangoClient.MustNewInCluster()
+
+	// Prepare deployment config
+	deploymentName := "test-upgrade-" + string(mode) + "-" + string(engine)
```
ewoutp: This looks a bit complicated.

`deploymentName := strings.Replace(fmt.Sprintf("test-upgrade-%s-%s-%sto%s", mode, engine, fromVersion, toVersion), ".", "-", -1)`

tests/test_util.go
```diff
@@ -297,3 +310,72 @@ func removeSecret(cli kubernetes.Interface, secretName, ns string) error {
 	}
 	return nil
 }
+
+func waitUntilArangoDeploymentHealthy(deployment *api.ArangoDeployment, DBClient driver.Client, k8sClient kubernetes.Interface, versionString string) error {
```
ewoutp: Still needs a comment

PR #81, opened 2018-03-27
AsOwner no longer things the owner refers to a controller. It refers to the ArangoDeployment

ewoutp:
This was incorrect anyway.
Fixing it makes the pods show up properly in the k8s dashboard.


ewoutp:
When you create a resource in k8s, you can attach it to an owner.
That way when you delete the owner, the things attached to it are also deleted.

The `AsOwner` function creates a proper owner reference. The problem was that it did set the Controller field which (according to the docs) means that the reference was to the controller (so our operator) instead of the ArangoDeployment itself. This was wrong, because the intent was to attach to the ArangoDeployment, not its controller.

PR #83, opened 2018-03-27
Deployment state -> phase

ewoutp:
This PR change the `State` field in `DeploymentStatus` to `Phase` and greatly reduced the number of phases.
The Phase is intended to indicate where in the lifetime a deployment is.
All other state related indicators are now communicated using conditions.

Furthermore, this PR prevents a deployment from going into a Failed phase (from which there is no recover possible) where-ever possible.


pkg/deployment/deployment.go
```diff
@@ -168,38 +168,33 @@ func (d *Deployment) send(ev *deploymentEvent) {
...
-			d.failOnError(err, "Failed to create secrets")
-			return
+			d.CreateEvent(k8sutil.NewErrorEvent("Failed to create secrets", err, d.GetAPIObject()))
 		}
 
 		// Create services
 		if err := d.resources.EnsureServices(); err != nil {
-			d.failOnError(err, "Failed to create services")
-			return
+			d.CreateEvent(k8sutil.NewErrorEvent("Failed to create services", err, d.GetAPIObject()))
```
neunhoef: Does `d.CreateEvent` somehow leave this function early (maybe with a panic)? If not, is it right to simply move on, for example if the services have not been setup, should we go on to pods?

ewoutp: CreateEvent just comes back.
The inspection loop will keep trying to create all resources, so if something fails here, we just keep trying later.

PR #92, opened 2018-03-29
Check member failure

ewoutp:
This PR adds a new high level layer of checks to improve resilience.
It introduced a phase "Failed" for a member. When a member has reached that phase, there is no hope of recovery and it will be removed. The existing reconciliation rules ensure that a new member will be added.

The resilience check goes over all members and checks for signs that the member is dead beyond hope of recovery. If so, it checks if the member is allowed to be replaced. If that is all the case, the member phase is set to failed. The reconciler will create a plan to remove it.


pkg/util/arangod/agency_health.go
```diff
@@ -0,0 +1,98 @@
...
+	LeaderEndpoint string
+	IsResponding   bool
+}
+
+// AreAgentsHealthy performs a health check on all given agents.
+// Of the given agents, 1 must respond as leader and all others must redirect to the leader.
+// The function returns nil when all agents are healthy or an error when something is wrong.
+func AreAgentsHealthy(ctx context.Context, clients []Agency) error {
+	wg := sync.WaitGroup{}
+	invalidKey := []string{"does-not-exists-149e97e8-4b81-5664-a8a8-9ba93881d64c"}
```
kvahed: Just for purpose of beauty:  drop the `s` in `exists`

ewoutp: done

PR #95, opened 2018-03-30
Check contents of persisted volume when dbserver is restarting

ewoutp:
This PR adds additional safety when restarting a dbserver.
It ensures that the persistent volume contains a UUID file with expected content
and an ENGINE file with expected content.


kvahed:
Is this functionality reflected in test? Do you find that a good idea for a simple test?

ewoutp:
Adding a test would be good, probably not simple

PR #96, opened 2018-03-30
Adding chaos-monkey for deployments

ewoutp:
This PR introduces a built-in chaos monkey for deployments.
The chaos monkey will kill a random pod at a specified interval with specified probability per interval.
To enable chaos, you must for allow it in the operator (`--chaos.allowed=true`) and then
enable it in the deployment specification.
```
  spec:
    chaos: 
      enabled: true
```

For development builds of the operator, chaos is allowed by default.
For release builds it is disallowed by default.


examples/simple-cluster.yaml
```diff
@@ -10,3 +10,5 @@ spec:
   coordinators:
     args:
       - --log.level=true
+  chaos:
```
kvahed: Redundant

ewoutp: Removed it

PR #108, opened 2018-04-04
Doc - Follow file name conventions of main docs, move to Tutorials

Simran-B:
Some changes as discussed with Santo.

Also see https://github.com/arangodb/arangodb/pull/5014

@ewoutp Is this a good general description of what kube-arangodb is?
https://github.com/arangodb/kube-arangodb/compare/docs-integration?expand=1#diff-f7213dcd3b2930325722b1d1627b0152

Should the file names of the design docs be changed as well to follow CamelCase convention as used by our main documentation?


docs/Manual/Deployment/Kubernetes/DeploymentResource.md
```diff
@@ -102,7 +102,7 @@ This setting specifies the type of storage engine used for all servers
 in the cluster.
 Possible values are:
 
-- `MMFiles` To use the MMfiles storage engine.
+- `MMFiles` To use the MMFiles storage engine.
```
ewoutp: I don't like this change. IMO it becomes confusing, since the possible value is case sensitive.
So even if the rest of the documentation uses `MMfiles`, I'll still prefer to keep `MMFiles` here.

docs/Manual/Deployment/Kubernetes/README.md
```diff
@@ -1,12 +1,6 @@
...
-- [Metrics](./metrics.md)
-- [Scaling](./scaling.md)
-- [Services & Load balancer](./services_and_loadbalancer.md)
-- [Storage](./storage.md)
-- [Storage Resource](./storage_resource.md)
-- [TLS](./tls.md)
-- [Upgrading](./upgrading.md)
+The ArangoDB Kubernetes Operator (`kube-arangodb`) is a set of two operators
+that you deploy in your Kubernetes cluster to manage deployments of the
+ArangoDB database and provide `PersistenVolumes` on local storage of your
```
ewoutp: `PersistentVolumes` (looks like we have that wrong in the entire document)

docs/Manual/Deployment/Kubernetes/Storage.md
```diff
@@ -1,7 +1,7 @@
 # Storage
 
 An ArangoDB cluster relies heavily on fast persistent storage.
-The ArangoDB operator uses `PersistenVolumeClaims` to deliver
+The ArangoDB Kubernetes Operator uses `PersistenVolumeClaims` to deliver
```
ewoutp: `PersistentVolumeClaims`

docs/Manual/Deployment/Kubernetes/README.md
Simran-B: Good catch, I didn't see the missing T.

docs/Manual/Deployment/Kubernetes/DeploymentResource.md
Simran-B: I changed it from MMfiles to MMFiles, isn't that what you prefer? In the API, the storage engine names are all lower case IIRC, but we use MMFiles and RocksDB in the docs in most places if it doesn't have to be verbatim.

ewoutp: Sorry, my mistake

PR #136, opened 2018-05-11
Feature: finalizers

ewoutp:
fixes #143 

This PR adds a finalizer to dbserver Pods and a preStop hook to prevent it from being deleted before all data is safe.
It also adds a finalizer to all PVCs to prevent them from being removed while their member still exists.

TODO:

- [x] Prevent orphaned resources
- [x] Add finalizer to agents to prevent removing to many at the same time
- [x] Run all tests


docs/design/lifecycle_hooks_and_finalizers.md
```diff
@@ -0,0 +1,37 @@
+# Lifecycle hooks & Finalizers
+
+The ArangoDB operator expects full control of the `Pods` and `PersistentVolumeClaims` it creates.
+Therefore it takes measures to prevent the removal of those resources
+until it is safe to do so.
+
+To achieve this, the server containers in the `Pods` have
+a `preStop` hook configured and finalizers are added to the `Pods`
+ands `PersistentVolumeClaims`.
```
neunhoef: "ands" -> "and"

lifecycle.go
```diff
@@ -0,0 +1,150 @@
...
+
+	if _, err := io.Copy(wr, rd); err != nil {
+		cliLog.Fatal().Err(err).Msg("Failed to copy")
+	}
+
+	// Set file mode
+	if err := os.Chmod(targetPath, 0755); err != nil {
+		cliLog.Fatal().Err(err).Msg("Failed to chmod")
+	}
+}
```
neunhoef: Just for the record (one more time): LOL about this trick.

pkg/deployment/resources/pod_finalizers.go
```diff
@@ -0,0 +1,179 @@
...
+		return nil
+	}
+	// Not cleaned out yet, check member status
+	if memberStatus.Conditions.IsTrue(api.ConditionTypeTerminated) {
+		log.Warn().Msg("Member is already terminated before it could be cleaned out. Not good, but removing drain dbserver finalizer because we cannot do anything further")
+		return nil
+	}
+	// Ensure the cleanout is triggered
+	log.Debug().Msg("Server is not yet clean out. Triggering a clean out now")
+	if err := cluster.CleanOutServer(ctx, memberStatus.ID); err != nil {
```
neunhoef: I cannot immediately see how often this is called. It seems that as long as the cleanout is ongoing, a new cleanout is ordered every time this routine is called, which happens every time `runPodFinalizers` is run. I found a loop doing this but I could not immediately determine how fast this loop could turn. I do not think it is a real problem to call CleanOutServer again, but I have a bad feeling if many of these calls would be done in rapid succession, whilst it is trying to clean itself out.

ewoutp: The interval depends on external events.
So while it is unlikely that these calls are made in rapid succession, it is possible.

PR #144, opened 2018-05-28
Fixes needed to run on latest openshift.

ewoutp:
This PR contains fixes needed to deploy ArangoDB on OpenShift.
Tested on `minishift v1.17.0+f974f0c`.

Note that in order to make it work, also a patched version of the ArangoDB image is needed.
A `Dockerfile` as below is known to work.

```
FROM arangodb/arangodb-preview:3.3
RUN chgrp 0 /var/lib/arangodb3 && chmod 775 /var/lib/arangodb3
RUN chgrp 0 /var/lib/arangodb3-apps && chmod 775 /var/lib/arangodb3-apps
```

These changes are patched into newly released arangodb images.

### DIFF
--- a/pkg/deployment/images.go
+++ b/pkg/deployment/images.go
@@ -167,6 +167,8 @@ func (ib *imagesBuilder) fetchArangoDBImageIDAndVersion(ctx context.Context, ima
 	args := []string{
 		"--server.authentication=false",
 		fmt.Sprintf("--server.endpoint=tcp://[::]:%d", k8sutil.ArangoPort),
+		"--database.directory=" + k8sutil.ArangodVolumeMountDir,
+		"--log.output=+",
 	}
 	terminationGracePeriod := time.Second * 30
 	tolerations := make([]v1.Toleration, 0, 2)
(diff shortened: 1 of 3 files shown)


pkg/deployment/images.go
```diff
@@ -167,6 +167,8 @@ func (ib *imagesBuilder) fetchArangoDBImageIDAndVersion(ctx context.Context, ima
 	args := []string{
 		"--server.authentication=false",
 		fmt.Sprintf("--server.endpoint=tcp://[::]:%d", k8sutil.ArangoPort),
```
matthewvon: does this imply that the connection will never be ssl?  and never use authentication?

ewoutp: For the ID pod : Yes.

ewoutp: The ID pod is run very quickly to fetch version information for a given image name.

ewoutp: It will never be accessible from outside, nor will it store any data or live longer than a few seconds

PR #147, opened 2018-05-31
Sync access packages

ewoutp:
This PR aims to make the (dc2dc) replication easier and such that it no longer requires `arangosync` to be installed locally.

It does so by:

- Simplify the settings in and `ArangoDeploymentReplication` (see docs)
- Add a concept of an `access package`. This is a k8s resource file containing a secret for the TLS CA and a secret with a generated client authentication keyfile. See docs for details.
The access package is created by editing the `ArangoDeployment` in the source cluster (add a name to the `accessPackageSecretNames` field). This yields the creation of a secret with that name.
Extract the secret data (under `accessPackage.yaml`).
Finally apply the `accessPackage.yaml` file in the destination cluster.


deps/github.com/arangodb/arangosync/client/client_cache.go
```diff
@@ -113,11 +116,13 @@ func NewAuthentication(tlsAuth TLSAuthentication, jwtSecret string) Authenticati
...
 	TLSAuthentication
 	JWTSecret string
+	Username  string
+	Password  string
 }
 
 // String returns a string used to unique identify the authentication settings.
 func (a Authentication) String() string {
-	return a.TLSAuthentication.String() + ":" + a.JWTSecret
+	return a.TLSAuthentication.String() + ":" + a.JWTSecret + ":" + a.Username + ":" + a.Password
```
matthewvon: I do not know where String() is used.  But I wonder if making the password available via this function is required, and if not required, is it a bad thing.

ewoutp: It is used to generate a key that matches the exact settings. So with exactly the same key you are allowed to share the same client object.

docs/Manual/Deployment/Kubernetes/DeploymentReplicationResource.md
```diff
@@ -103,8 +146,62 @@ the TLS connection created by the SyncMaster at the specified destination.
...
+
+```bash
+kubectl apply -f accessPackage.yaml
+```
 
-This setting specifies the name of a `Secret` containing a client authentication certificate,
-used to authenticate the SyncMaster in the destination cluster with the SyncMaster in the
-source cluster.
+As a result, the destination Kubernetes cluster will have 2 additional `Secrets`. One containing a client authentication certificate
+formatted ad a keyfile. Another containing the public key of the TLS CA certificate of the source cluster.
```
matthewvon: spelling or grammar or typo:  "One containing a client authentication certificate formatted ad a keyfile."  ad -> and ?  If so, better wording would be "One contains a ...".  Second sentence would then be "Another contains ..."

docs/Manual/Deployment/Kubernetes/DeploymentReplicationResource.md
```diff
@@ -103,8 +146,62 @@ the TLS connection created by the SyncMaster at the specified destination.
...
+- A TLS certificate authority public key, wrapped in a `Secret` in a `ca.crt` data field.
+
+The format of the access package is such that it can be inserted into a Kubernetes cluster using the standard `kubectl` tool.
+
+To create an access package that can be used to authenticate with the ArangoDB SyncMasters of an `ArangoDeployment`,
+add a name of a non-existing `Secret` to the `spec.sync.externalAccess.accessPackageSecretNames` field of the `ArangoDeployment`.
+In response, a `Secret` is created in that Kubernetes cluster, with the given name, that contains a `accessPackage.yaml` data field
+that contains a Kubernetes resource specification that can be inserted into the other Kubernetes cluster.
+
+The process for creating and using an access package to authentication at the source cluster is as follows:
```
matthewvon: "package to authentication"  -> "package for authentication"

PR #155, opened 2018-06-05
Operator high-availability

ewoutp:
This PR sets the number of replicas for the all operator to 2 (from 1). This ensures that 1 replica is always standing by (waiting in leader election) and able to take over quickly when the leader is not responding.
This is faster than waiting for kubernetes to terminate a pod that becomes unresponsive.

This PR also adds tolerations that were still missing on the deployment-replication operator.


manifests/templates/deployment-replication/deployment-replication.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   name: {{ .DeploymentReplication.OperatorDeploymentName }}
   namespace: {{ .DeploymentReplication.Operator.Namespace }}
 spec:
-  replicas: 1
+  replicas: 2
```
matthewvon: personal curiosity:  why 2 ?

PR #162, opened 2018-06-07
Force volumes to unique nodes for production environments

ewoutp:


pkg/storage/pv_creator.go
```diff
@@ -204,3 +246,96 @@ func createNodeAffinity(nodeName string) (string, error) {
 	}
 	return string(encoded), nil
 }
+
+// createNodeClientMap creates a map from node name to API.
+// Clients that do not respond properly on a GetNodeInfo request are
+// ignored.
+func createNodeClientMap(ctx context.Context, clients []provisioner.API) map[string]provisioner.API {
```
matthewvon: should I be harassing you about these functions being nicely isolated and yet have no unit tests to demonstrate they work?

PR #166, opened 2018-06-08
Added unit tests for pv_creator.go

ewoutp:
Implements https://github.com/arangodb/kube-arangodb/pull/162#discussion_r193785282

### DIFF
--- a/pkg/storage/pv_creator_test.go
+++ b/pkg/storage/pv_creator_test.go
@@ -99,8 +99,8 @@ func TestCreateNodeAffinity(t *testing.T) {
 // TestCreateNodeClientMap tests createNodeClientMap.
 func TestCreateNodeClientMap(t *testing.T) {
 	GB := int64(1024 * 1024 * 1024)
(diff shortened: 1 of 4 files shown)


pkg/storage/pv_creator_test.go
```diff
@@ -0,0 +1,209 @@
...
+		assert.NoError(t, err)
+		assert.Equal(t, expected, output, "Input: '%s'", input)
+	}
+}
+
+// TestCreateNodeClientMap tests createNodeClientMap.
+func TestCreateNodeClientMap(t *testing.T) {
+	GB := int64(1024 * 1024 * 1024)
+	foo := mocks.NewProvisioner("foo", 100*GB, 100*GB)
+	bar := mocks.NewProvisioner("bar", 100*GB, 100*GB)
```
matthewvon: trivial:  my preference would be to have four different values instead of 100*GB for all.  What if you happen to have one of parameters backwards some place in the code?

PR #167, opened 2018-06-08
Various test improvements

ewoutp:

### DIFF
--- a/tests/cursor_test.go
+++ b/tests/cursor_test.go
@@ -29,7 +29,7 @@ import (
 	"time"
 
 	"github.com/dchest/uniuri"
-	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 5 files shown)


tests/cursor_test.go
```diff
@@ -29,7 +29,7 @@ import (
 	"time"
 
 	"github.com/dchest/uniuri"
-	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
neunhoef: In this file you switch from "assert" to "require", in all the others not. Why is that?

ewoutp: In the cases where `require` is used, other tests follow it.
I don't want to run these following tests when the requirement has failed

PR #171, opened 2018-06-12
Abort reconcilientation plan on failed cleanout server

ewoutp:
This PR checks the status of ongoing cleanout server action by inspecting the right keys in the agency.
If a cleanout server job has failed, the reconciliation plan is aborted.

This PR uses https://github.com/arangodb/go-driver/pull/131

### DIFF
--- a/pkg/util/arangod/cleanout_server.go
+++ b/pkg/util/arangod/cleanout_server.go
@@ -1,0 +1,99 @@
+//
+// DISCLAIMER
+//
+// Copyright 2018 ArangoDB GmbH, Cologne, Germany
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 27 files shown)


pkg/util/arangod/cleanout_server.go
```diff
@@ -0,0 +1,99 @@
...
+// String returns a string representation of the given state.
+func (s CleanoutJobStatus) String() string {
+	return fmt.Sprintf("state: '%s', reason: '%s'", s.state, s.reason)
+}
+
+var (
+	agencyJobStateKeyPrefixes = [][]string{
+		{"arango", "Target", "ToDo"},
+		{"arango", "Target", "Pending"},
+		{"arango", "Target", "Done"},
```
neunhoef: It is `Finished` instead of `Done`. I probably told you wrongly.

pkg/util/arangod/cleanout_server.go
```diff
@@ -0,0 +1,99 @@
...
+	reason string
+}
+
+// IsFailed returns true when the job is failed
+func (s CleanoutJobStatus) IsFailed() bool {
+	return s.state == "Failed"
+}
+
+// IsDone returns true when the job is finished
+func (s CleanoutJobStatus) IsDone() bool {
```
neunhoef: Should be called `Finished` here as well, see below.

ewoutp: Changed Done to Finished & `IsDone` to `IsFinished` (which is not being used)

PR #172, opened 2018-06-12
Avoid overwriting status changes

ewoutp:
There were some cases where status could be overwritten, resulting in member objects being listed twice.
This PR makes modifying the status a very controlled procedure. When you fetch the status you get a last version number. When updating the status, that last version has to match the current version number.

### DIFF
--- a/pkg/deployment/context_impl.go
+++ b/pkg/deployment/context_impl.go
@@ -24,6 +24,8 @@ package deployment
 
 import (
 	"context"
+	"fmt"
+	"sync/atomic"
 
 	"github.com/arangodb/arangosync/client"
 	"github.com/arangodb/arangosync/tasks"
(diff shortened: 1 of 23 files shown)


pkg/deployment/context_impl.go
```diff
@@ -63,20 +65,37 @@ func (d *Deployment) GetNamespace() string {
...
 	return d.apiObject.Spec
 }
 
 // GetStatus returns the current status of the deployment
-func (d *Deployment) GetStatus() api.DeploymentStatus {
-	return d.status
+// together with the current version of that status.
+func (d *Deployment) GetStatus() (api.DeploymentStatus, int32) {
+	version := atomic.LoadInt32(&d.status.version)
+	return *d.status.last.DeepCopy(), version
```
neunhoef: What if `*d.status.last` (and indeed `*d.status.version`) have changed between the execution of line 81 and line 82? Do we not have to check the version again after the DeepCopy?
Or we have to make getting the version and doing a DeepCopy an atomic operation (with a Mutex)...

pkg/deployment/context_impl.go
```diff
@@ -63,20 +65,37 @@ func (d *Deployment) GetNamespace() string {
...
+func (d *Deployment) UpdateStatus(status api.DeploymentStatus, lastVersion int32, force ...bool) error {
+	if !atomic.CompareAndSwapInt32(&d.status.version, lastVersion, lastVersion+1) {
+		// Status is obsolete
+		d.deps.Log.Error().
+			Int32("expected-version", lastVersion).
+			Int32("actual-version", d.status.version).
+			Msg("UpdateStatus version conflict error.")
+		return maskAny(fmt.Errorf("Status conflict error. Expected version %d, got %d", lastVersion, d.status.version))
+	}
+	d.status.last = *status.DeepCopy()
```
neunhoef: Again: What happens if somebody updates version and status between the execution of lines 90 and 98?

pkg/deployment/reconcile/plan_executor.go
```diff
@@ -96,14 +98,15 @@ func (d *Reconciler) ExecutePlan(ctx context.Context) (bool, error) {
...
 			if ready {
-				// action.CheckProgress may have changed status, so reload it.
-				status = d.context.GetStatus()
-				// Remove action from list
-				status.Plan = status.Plan[1:]
-				// Save plan update
-				if err := d.context.UpdateStatus(status); err != nil {
-					log.Debug().Err(err).Msg("Failed to update CR status")
-					return false, maskAny(err)
+				{ // action.CheckProgress may have changed status, so reload it.
```
neunhoef: What does this extra pair of braces achieve?

ewoutp: It provides an scope in which we're sure we're using the right `status`.

ewoutp:
The point of this PR is NOT to make it go-routine safe.
It wants to prevent the following:
- Fetch status into variable A 
- do something, changing a field of A
- Fetch status into variable B

neunhoef:
Understood, having another quick look.

pkg/deployment/context_impl.go
```diff
@@ -63,20 +64,44 @@ func (d *Deployment) GetNamespace() string {
...
+		d.deps.Log.Error().
+			Int32("expected-version", lastVersion).
+			Int32("actual-version", d.status.version).
+			Msg("UpdateStatus version conflict error.")
+		return maskAny(fmt.Errorf("Status conflict error. Expected version %d, got %d", lastVersion, d.status.version))
+	}
+	d.status.version++
+	d.status.last = *status.DeepCopy()
 	if err := d.updateCRStatus(force...); err != nil {
 		return maskAny(err)
```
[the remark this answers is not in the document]
ewoutp: If there is a retry higher up, it MUST call `GetStatus()` first to get the latest status and version.
So while a network error will cause a state diff between in-memory & persistent in kube-apiserver, this will not affect future updates.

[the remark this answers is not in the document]
neunhoef: ok

PR #201, opened 2018-06-26
All operator Pods will now reach the Ready state.

### LINKED ISSUE #198: Fix operators Deployment spec wrt minimum availability
Current settings complain about "Not having the minimum availability"

ewoutp:
This PR changes the behavior of the Ready state of the Pods that run the operators.

Before such a Pod was only marked Ready when it has won the leader election.
In this PR, such a Pod is marked Ready as soon as a new leader has been detected.

Other changes are:
- Leader election lifecycle adds events to the Deployment that owns the operator.
- Pod is exited with exitcode 1 when Pod that won the leader election at some point cannot renew its leader status.

fixes #198


pkg/operator/operator_leader.go
```diff
@@ -23,16 +23,37 @@
...
+
+	"github.com/arangodb/kube-arangodb/pkg/util/k8sutil"
+	"github.com/arangodb/kube-arangodb/pkg/util/probe"
 )
 
-func (o *Operator) runLeaderElection(lockName string, onStart func(stop <-chan struct{})) {
+// runLeaderElection performs a leader election on a lock with given name in
+// the namespace that the operator is deployed in.
+// When the leader election is won, the given callback is called.
+// When the leader election is lost (even after it was won once), the process is killed.
```
neunhoef: This comment confused me. If a new pod tries to get elected and "loses" the election, then the process remains lingering around and will try again. It is only if *leadership* is lost that the process exits. Please adjust the comment. Code is OK, as far as I see.

ewoutp: done

PR #205, opened 2018-06-27
Documentation fixes

Simran-B:
Only typos so far, might have to fix a few Gitbook related issues as well.


Simran-B:
We won't add Cygwin-specific workarounds for the template render issues at this time as discussed with @sleto-it. I'll try to get documentation building working with the Windows Subsystem for Linux (WSL). We should edit the Cookbook and Documentation maintainer readme accordingly if we don't support Cygwin anymore. One final test could be to use LF line endings for everything in git and re-clone to make sure it's not a line ending issue again.

PR #211, opened 2018-07-02
Feature/dashboard

ewoutp:
This PR lays the groundwork for a dashboard for all operators in kube-arangodb.

This PR implements some of the functionality of the ArangoDeployment operator dashboard.
More functionality will be added in later PRs.
Other operators (replication & storage) will also be handled in other PRs.


dashboard/src/App.js
```diff
@@ -0,0 +1,47 @@
...
+    this.reloadOperators();
+  }
+
+  async reloadOperators() {
+    const operators = await apiGet('/api/operators');
+    this.setState({operators});
+  }
+
+  render() {
+    setTimeout(this.reloadOperators.bind(this), 5000);
```
pluma: This should happen in componentDidMount. Having side-effects in a render function is generally a bad idea.

dashboard/src/App.js
```diff
@@ -0,0 +1,47 @@
...
+    let children = [];
+    if (this.state.operators.deployment) {
+      children.push((<DeploymentOperator/>));
+    } else {
+      children.push((<NoOperator/>));
+    }
+    return (
+      <Container>
+        {children}
```
pluma: This could be inlined as `{this.state.operators.deployment ? <DeploymentOperator/> : <NoOperator/>}`.

dashboard/src/App.js
```diff
@@ -0,0 +1,47 @@
...
+        <Segment basic>
+          <Message>
+            <Message.Header>Kube-ArangoDB</Message.Header>
+            <p>
+              Running in Pod <b>{this.state.operators.pod}</b> in namespace <b>{this.state.operators.namespace}</b>.
```
pluma: These will be `undefined` until the request in `componentDidMount` resolves. React will render those variables as empty strings in this case but that'll look a bit ugly.

dashboard/src/App.js
```diff
@@ -0,0 +1,47 @@
...
+            </p>
+          </Message>     
+        </Segment>
+      </Container>
+    );
```
pluma: TBH, I'd probably extract most of the render function into a separate component that's entirely presentational, e.g.

```jsx
const OperatorsView = ({deployment, pod, namespace}) => (
  <Container>
    {deployment ? <DeploymentOperator /> : <NoOperator />}
    <Segment basic>
      <Message>
        <Message.Header>Kube-ArangoDB</Message.Header>
        <p>Running in Pod <b>{pod}</b> in namespace <b>{namespace}</b>.</p>
      </Message>
    </Segment>
  </Container>
));
class App extends Component {
  state = {}
  componentDidMount () {
    this.reloadOperators();
    this._interval = setInterval(this.reloadOperators, 5000);
  }
  componentWillUnmount () {
    clearInterval(this._interval);
  }
  reloadOperators = async () => {
    const operators = await apiGet('...');
    this.setState({ operators });
  }
  render () {
    if (this.state.operators) {
      return <OperatorsView {...this.state.operators} />;
  }
  return <div>Loading...</div>;
}
```

Keep in mind this still doesn't account for timeouts or errors in the async stuff.

dashboard/src/api/api.js
```diff
@@ -0,0 +1,8 @@
+// apiGet performs a GET request on the API with given local URL.
+// The result is decoded from JSON and returned.
+export async function apiGet(localURL: string): Promise<Any> {
```
pluma: Here you're using TypeScript syntax but the files are JS files.

If you're going to use Create React App you probably want to use it with the TypeScript react-scripts:
https://github.com/Microsoft/TypeScript-React-Starter#create-our-new-project

dashboard/src/deployment/DeploymentList.js
```diff
@@ -0,0 +1,49 @@
...
+    this.reloadDeployments();
+  }
+
+  async reloadDeployments() {
+    const result = await apiGet('/api/deployment');
+    this.setState({items:result.deployments});
+  }
+
+  render() {
+    setTimeout(this.reloadDeployments.bind(this), 5000);
```
pluma: Again see above about side-effects.

dashboard/src/App.js
pluma: Also I think you want to use setInterval as setTimeout is for one-off delayed code.

pluma: Also keep in mind setTimeout/setInterval will still keep firing if the component is unmounted. So you need to clean those up in componentWillUnmount.

dashboard/src/deployment/DeploymentList.js
```diff
@@ -0,0 +1,49 @@
...
+    const items = this.state.items;
+    if (items.length === 0) {
+      return (<div>No deployments</div>);
+    }
+    return (
+      <Table striped celled>
```
pluma: Again, I'd like to keep the presentational part in a presentational component to make it easier to swap out later.

dashboard/src/deployment/DeploymentRow.js
```diff
@@ -0,0 +1,19 @@
...
+import { Icon, Table } from 'semantic-ui-react';
+//import logo from './logo.svg';
+//import './App.css';
+
+class DeploymentRow extends Component {
+  render() {
+    return (
+      <Table.Row>
+        <Table.Cell><Icon name="bell" color="red"/></Table.Cell>
+        <Table.Cell>{this.props.info.name}</Table.Cell>
```
pluma: No reason to use a class based component unless you have some kind of state. Also I'd unpack the info properties into props directly.

dashboard/src/App.js
pluma: FYI the `{...this.state.operators}` part is just me being lazy because I'm coding in the browser. I'd unpack the props explicitly and pass them individually. The `reloadOperators = async () => {...}` stuff binds the function to the instance so you don't need to bind it explicitly when creating the timeout/interval. This is generally a good practice when dealing with things like event handlers but not entirely necessary here (just easier to avoid mistakes by doing it here too).

dashboard/src/deployment/DeploymentList.js
```diff
@@ -1,48 +1,67 @@
...
+    <Table.Row>
+      <Table.HeaderCell>State</Table.HeaderCell>
+      <Table.HeaderCell>Name</Table.HeaderCell>
+      <Table.HeaderCell>Mode</Table.HeaderCell>
+      <Table.HeaderCell>Pods</Table.HeaderCell>
+    </Table.Row>
+  </Table.Header>
+);
+
+const RowView = ({info}) => (
```
pluma: I'd pass in the item's properties individually to make the props more explicit here.

dashboard/src/deployment/DeploymentList.js
```diff
@@ -1,48 +1,67 @@
...
+    <Table.Cell>{info.ready_pod_count} / {info.pod_count}</Table.Cell>
+  </Table.Row>
+);
+
+const ListView = ({items}) => (
+  <Table striped celled>
+    <HeaderView/>
+    <Table.Body>
+      {
+        (items) ? items.map((item) => <RowView key={item.name} info={item}/>) : <p>No items</p>
```
pluma: i.e. here I'd do `<RowView key={item.name} name={item.name} mode={item.mode} readyCount={item.ready_pod_count} totalCount={item.pod_count} />` so it's clearer what this takes.

dashboard/src/util/Loading.js
```diff
@@ -0,0 +1,17 @@
+import React, { Component } from 'react';
+import { Dimmer, Loader, Segment } from 'semantic-ui-react';
+
+class Loading extends Component {
+  render() {
```
pluma: Again, if this doesn't need state or anything special, just use a function component here. Also helps increasing the visibility of what props it takes.

dashboard/src/api/api.js
pluma: BTW if you do use this, make sure to simplify the `extends` list in the tslint.json it generates:
```
"extends": ["tslint-config-prettier"],
```
The defaults are overly pedantic and tend to get in the way.

dashboard/src/App.js
```diff
@@ -6,16 +6,19 @@ import { apiGet } from './api/api.js';
...
-    {deployment ? <DeploymentOperator /> : <NoOperator />}
-    <Segment basic>
-      <Message>
-        <Message.Header>Kube-ArangoDB</Message.Header>
-        <p>Running in Pod <b>{pod}</b> in namespace <b>{namespace}</b>.</p>
-      </Message>
-    </Segment>
-  </Container>
+  <div>
+    {deployment ? <DeploymentOperator pod-info={<PodInfoView pod={pod} namespace={namespace}/>}/> : <NoOperator />}
```
pluma: BTW React components normally use `camelCase` for prop names instead of `kebab-case` as HTML does.

dashboard/src/deployment/DeploymentDetails.js
```diff
@@ -0,0 +1,88 @@
+import React, { Component } from 'react';
+import { apiGet } from '../api/api.js';
+import { Accordion, Header, Icon, List, Segment } from 'semantic-ui-react';
+import Loading from '../util/Loading.js';
+//import CommandInstruction from '../util/CommandInstruction.js';
+
+const MemberGroupsView = ({member_groups}) => (
```
pluma: Here too `camelCase` would be more idiomatic.

dashboard/src/deployment/DeploymentDetails.js
```diff
@@ -0,0 +1,88 @@
...
+      members={item.members}
+    />)}
+  </div>
+);
+
+const MemberListView = ({group, activeMemberID, onClick, members}) => (
+  <Segment>
+    <Header>{group}</Header>
+    <List divided>
+      {members.map((item) => <MemberView memberInfo={item} active={item.id === activeMemberID} onClick={onClick}/>)}
```
pluma: May want to add `key={item.id}` to help React diffing renders when the members prop changes.

dashboard/src/deployment/DeploymentDetails.js
```diff
@@ -0,0 +1,88 @@
...
+      </Accordion.Content>
+    </Accordion>
+  </List.Item>
+);
+
+class MemberListComponent extends Component {
+  state = {};
+
+  onClick = (id) => { 
+    this.setState({activeMemberID:(this.state.activeMemberID === id) ? null : id});
```
pluma: When using `this.state` to set state, you should use the functional form of `setState`:

```js
this.setState(({activeMemberID}) => ({
  activeMemberID: activeMemberID === id ? null : id
}));
```

Using `this.state` to derive new state is error-prone as `setState` is potentially async.

pluma: BTW, I'd call the method "handleClick" so you pass `onClick={this.handleClick}` but I honestly have no idea where that convention comes from.

dashboard/src/deployment/DeploymentList.js
```diff
@@ -28,14 +80,34 @@ const ListView = ({items}) => (
...
+          />) : <p>No items</p>
       }
     </Table.Body>
   </Table>
 );
 
 const EmptyView = () => (<div>No deployments</div>);
 
+function createDeleteCommand(name, namespace) {
+  return `kubectl delete ArangoDeployment -n ${namespace} ${name}`;
```
pluma: If you're not using this in multiple places I'd just inline it.

dashboard/src/util/CommandInstruction.js
```diff
@@ -0,0 +1,40 @@
+import React, { Component } from 'react';
+import { Button, Modal, Segment } from 'semantic-ui-react';
+import {CopyToClipboard} from 'react-copy-to-clipboard';
+
+class CommandInstruction extends Component {
+  state = {};
```
pluma: In this case I'd initialise `state = {open: false}` explicitly because having it `undefined` doesn't really add any meaning distinct from `false`.

dashboard/src/util/Loading.js
```diff
@@ -1,17 +1,13 @@
...
-  render() {
-    return (
+const Loading = ({message}) => (
         <Segment>
         <Dimmer inverted active>
-          <Loader inverted>{this.props.message || "Loading..."}</Loader>
+          <Loader inverted>{message || "Loading..."}</Loader>
         </Dimmer>
-        <div style={{"min-height":"3em"}}/>
+        <div style={{minHeight:"3em"}}/>
```
pluma: For consistency you might want to use `styled` for this one. Remember emotion also provides a `css` template handler which simply generates a classname if you don't want to keep track of another component just for the CSS class.

dashboard/src/auth/Auth.js
```diff
@@ -0,0 +1,82 @@
...
+
+  async componentDidMount() {
+    try {
+      api.token = this.state.token;
+      await api.get('/api/operators');
+      this.setState({
+        authenticated: true,
+        showLoading: false,
+        token: api.token
+      });
```
pluma: If I understand this right, you're setting the global `api` instance's `token` property to the one from state, then doing a call and discarding the result, then setting the state's `token` to that of the `api`.

Does the `token` change during the request? Does the request have side-effects? I'm guessing you only use the request to determine whether you're authenticated and authorized.

dashboard/src/api/api.js
```diff
@@ -0,0 +1,49 @@
...
+    // apiGet performs a GET request on the API with given local URL.
+    // The result is decoded from JSON and returned.
+    async get(localURL) {
+        let headers = {
+            'Accept': 'application/json'
+        };
+        if (this.token) {
+            headers['Authorization'] = `bearer ${this.token}`; 
+        }
+        const result = await fetch(localURL, {headers});
```
pluma: Just an FYI because fetch can be a bit too low-level sometimes and have cross-browser surprises: https://yarnpkg.com/en/package/axios

dashboard/src/auth/Login.js
```diff
@@ -0,0 +1,54 @@
+import React, { Component } from 'react';
+import { Button, Container, Form, Icon, Message, Modal } from 'semantic-ui-react';
+
+const LoginView = ({username, password, usernameChanged, passwordChanged, doLogin, error}) => (
+  <Container>
+    <Form onSubmit={doLogin}>
+      <Form.Field>
+        <label>Name</label>
+        <input focus="true" value={username} onChange={(e) => usernameChanged(e.target.value)}/>
```
pluma: Conventionally this would be called "onUsernameChange" I think. This also helps indicate it's an event handler rather than e.g. a flag.

dashboard/src/deployment/DeploymentDetails.js
```diff
@@ -0,0 +1,72 @@
...
+  margin-right: 1em;
+  margin-top: 1em;
+  max-width: 0;
+  display: inline-block;
+`;
+
+const MemberGroupsView = ({memberGroups, namespace}) => (
+  <div>
+    {memberGroups.map((item) => <MemberList 
+      key={`server-group-${item.group}`}
```
pluma: Any particular reason you're prefixing the keys? They're local to their parent so this doesn't seem to do much.

dashboard/src/deployment/DeploymentList.js
```diff
@@ -0,0 +1,191 @@
...
+    case "red":
+      return "The deployment is in a bad state and manual intervention is likely needed.";
+    default:
+      return "State is not known.";
+  }
+}
+
+class DeploymentList extends Component {
+  state = {
+    items: undefined,
```
pluma: If you need an explicit "this is a thing but it's not set yet" value, `null` is generally preferable to `undefined` as it's difficult to distinguish values that exist but are `undefined` form values that are undefined because they don't exist.

dashboard/src/auth/Auth.js
pluma: I'm not really happy with the `Auth` component and the `LogoutContext` but I can't think of an easy way to improve on this either, so lgtm :man_shrugging:

dashboard/src/util/Loading.js
```diff
@@ -0,0 +1,13 @@
...
+import { Dimmer, Loader, Segment } from 'semantic-ui-react';
+
+const Loading = ({message}) => (
+        <Segment>
+        <Dimmer inverted active>
+          <Loader inverted>{message || "Loading..."}</Loader>
+        </Dimmer>
+        <div style={{minHeight:"3em"}}/>
+      </Segment>
+      );
```
pluma: BTW the indenting is a bit messy. Have you heard of our lord and saviour, Prettier? https://prettier.io/

pluma:
None of my comments are blocking. :shipit: if @neunhoef approves

dashboard/src/App.js
```diff
@@ -0,0 +1,66 @@
+import React, { Component } from 'react';
+import ReactTimeout from 'react-timeout';
+import DeploymentOperator from './deployment/DeploymentOperator.js';
+import NoOperator from './NoOperator.js';
+import Loading from './util/Loading.js';
+import api from './api/api.js';
+import { Container, Segment, Message } from 'semantic-ui-react';
+import './App.css';
```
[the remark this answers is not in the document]
ewoutp: they are (already) being removed in the next PR

dashboard/src/NoOperator.js
```diff
@@ -0,0 +1,24 @@
+import React, { Component } from 'react';
+import logo from './logo.svg';
+import './App.css';
```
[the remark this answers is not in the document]
ewoutp: plain css will be gone

dashboard/src/auth/Auth.js
ewoutp: correct.

dashboard/src/deployment/DeploymentDetails.js
ewoutp: fixed

dashboard/src/auth/Login.js
ewoutp: fixed

dashboard/src/deployment/DeploymentList.js
ewoutp: fixed

PR #212, opened 2018-07-05
Adjust documentation based on new load balancer support.

danielhlarkin:
The load balancer support is currently in devel in the main repo. It will be back-ported to 3.3 for sure, but has not been done yet. Do the docs here apply for 3.3? 3.2?

### DIFF
--- a/docs/Manual/Deployment/Kubernetes/DriverConfiguration.md
+++ b/docs/Manual/Deployment/Kubernetes/DriverConfiguration.md
@@ -116,10 +116,10 @@ follow-up request.
 
 ArangoDB will transparently forward any mismatched requests to the correct
 coordinator, so the requests can be answered correctly without any additional
-configuration. However, this incurs a small performance penalty due to the extra
+configuration. However, this incurs a small latency penalty due to the extra
 request across the internal network.
 
-To resolve this uncertainty client-side, make sure to run your client
+To prevent this uncertainty client-side, make sure to run your client
 application in the same Kubernetes cluster and synchronize your endpoints before
 making the initial query request. This will result in the use (by the driver) of
 internal DNS names of all coordinators. A follow-up request can then be sent to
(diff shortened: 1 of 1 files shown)


ewoutp:
Can't spot any differences other than additional newlines.

docs/Manual/Deployment/Kubernetes/DriverConfiguration.md
```diff
@@ -93,36 +93,42 @@ This results in a file called `ca.crt` containing a PEM encoded, x509 CA certifi
...
+requests must be handled by the same coordinator to which the initial request
+was made. As soon as there is a load-balancer between your client application
+and the ArangoDB cluster, it is uncertain which coordinator will receive the
+follow-up request.
+
+ArangoDB will transparently forward any mismatched requests to the correct
+coordinator, so the requests can be answered correctly without any additional
+configuration. However, this incurs a small performance penalty due to the extra
+request across the internal network.
+
```
danielhlarkin: @ewoutp The paragraph above is the main change.

ewoutp: Ok. We should also adjust the 2 paragraphs below, since they will no longer be needed, unless you want to avoid a small latency penalty.

docs/Manual/Deployment/Kubernetes/DriverConfiguration.md
```diff
@@ -93,36 +93,42 @@ This results in a file called `ca.crt` containing a PEM encoded, x509 CA certifi
...
+Follow-up requests have to be made to fetch the remaining data. These follow-up
+requests must be handled by the same coordinator to which the initial request
+was made. As soon as there is a load-balancer between your client application
+and the ArangoDB cluster, it is uncertain which coordinator will receive the
+follow-up request.
+
+ArangoDB will transparently forward any mismatched requests to the correct
+coordinator, so the requests can be answered correctly without any additional
+configuration. However, this incurs a small performance penalty due to the extra
```
ewoutp: Suggest to replace performance by latency

PR #213, opened 2018-07-09
Dashboard for ArangoLocalStorage operator

ewoutp:
This PR adds dashboard functionality to the ArangoLocalStorage operator.

### DIFF
--- a/dashboard/src/storage/StorageList.js
+++ b/dashboard/src/storage/StorageList.js
@@ -1,12 +1,11 @@
-import { Accordion, Header, Icon, Loader, Popup, Table } from 'semantic-ui-react';
+import { Loader, Table } from 'semantic-ui-react';
 import api, { isUnauthorized } from '../api/api.js';
-import CommandInstruction from '../util/CommandInstruction.js';
-import VolumeList from './VolumeList.js';
 import Loading from '../util/Loading.js';
 import React, { Component } from 'react';
 import ReactTimeout from 'react-timeout';
 import styled from 'react-emotion';
 import { withAuth } from '../auth/Auth.js';
+import StorageRow from './StorageRow';
 
 const LoaderBox = styled('span')`
   float: right;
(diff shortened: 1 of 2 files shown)


dashboard/src/api/api.js
```diff
@@ -1,13 +1,22 @@
+export function IsUnauthorized(e) {
```
pluma: FYI in most JS styles only classes and components use leading uppercase. So this should be `isUnauthorized`.

dashboard/src/App.js
```diff
@@ -16,11 +17,14 @@ const PodInfoView = ({pod, namespace}) => (
...
 
-const OperatorsView = ({error, deployment, pod, namespace}) => {
+const OperatorsView = ({error, deployment, storage, pod, namespace}) => {
   const podInfoView = (<PodInfoView pod={pod} namespace={namespace}/>);
   if (deployment) {
     return (<DeploymentOperator podInfoView={podInfoView} error={error}/>);
   }
+  if (storage) {
+    return (<StorageOperator podInfoView={podInfoView} error={error}/>);
+  }
```
pluma: If you want to avoid repeating yourself a bit, you could do something like this:

```js
let Operator = NoOperator;
if (deployment) Operator = DeploymentOperator;
else if (storage) Operator = StorageOperator;
return (
  <Operator
    podInfoView={
      <PodInfoView pod={pod} namespace={namespace} />
    }
    error={error}
  />
);
```

dashboard/src/App.js
```diff
@@ -43,24 +47,33 @@ class App extends Component {
...
+        pod={this.state.operators.pod}
+        namespace={this.state.operators.namespace}
+        />;
     }
     return (<LoadingView/>);
   }
 }
 
-export default ReactTimeout(App);
+export default ReactTimeout(withAuth(App));
```
pluma: FYI this is why HOCs have been losing some traction to render props recently: you're now injecting props from two places. This is probably fine for the time being but it can get a bit confusing as the complexity grows.

dashboard/src/NoOperator.js
```diff
@@ -1,22 +1,23 @@
...
-            <img src={logo} className="App-logo" alt="logo" />
-            <h1 className="App-title">Welcome to Kube-ArangoDB</h1>
-          </header>
-          <p className="App-intro">
-            There are no operators available yet.
-          </p>
-          {this.props.podInfoView}
-          {(this.props.error) ? <Message error content={this.props.error}/> : null}
-        </div>
+        <Container>
```
pluma: This `NoOperator` component is exactly what I meant when I talked about presentational components btw. It doesn't have any state and it only contains display logic (i.e. "if there's an error message, put it here"). If you want to make it easier to swap out the styling later, you should try to extract all the actual presentation stuff into components like this one.

Minor nitpick: no point in using a class component here, you could just make it a function.

dashboard/src/auth/Login.js
```diff
@@ -12,6 +13,7 @@ const LoginView = ({username, password, onUsernameChanged, onPasswordChanged, do
         <label>Password</label>
         <input type="password" value={password} onChange={(e) => onPasswordChanged(e.target.value)}/>
       </Form.Field>
+      <Form.Button className={css`display:none`} type="submit" />
```
pluma: Any particular reason you have a hidden submit button here? Pressing enter in the input field should already trigger the form submit.

dashboard/src/storage/StorageList.js
```diff
@@ -0,0 +1,208 @@
...
+    if (!items) {
+      return (<Loading />);
+    }
+    if (items.length === 0) {
+      return (<EmptyView />);
+    }
+    return (<ListView items={items} loading={this.state.loading} />);
+  }
+}
+
```
pluma: This file is getting rather large. I'd split it out to a folder. I generally aim to have one component per folder but YMMV.

dashboard/src/auth/Login.js
pluma: Also curious why you're using `css` here but `styled` elsewhere as most uses of `styled` can also be replaced with `css` (and generally vice versa unless you want to apply the same class to different components, in which case you'll always need `css`).

dashboard/src/api/api.js
ewoutp: changed

dashboard/src/App.js
ewoutp: changed

dashboard/src/auth/Login.js
ewoutp: hidden button was out of need. I expected it to work without, but it did not.

ewoutp: I like `styled` better, but that did not work in combination with `Form.Button`.

dashboard/src/storage/StorageList.js
ewoutp: once all the basic are in, I'll go over all of it for issues like this.

PR #218, opened 2018-07-11
Load balancing tests

danielhlarkin:
Needs significant testing and probably fixes, additional code.

This PR now imports https://github.com/arangodb/go-driver/pull/139


deps/github.com/arangodb/go-driver/context.go
```diff
@@ -52,11 +52,12 @@ const (
 	keyIsSystem                 ContextKey = "arangodb-isSystem"
 	keyIgnoreRevs               ContextKey = "arangodb-ignoreRevs"
 	keyEnforceReplicationFactor ContextKey = "arangodb-enforceReplicationFactor"
-	keyConfigured               ContextKey = "arangodb-configured"
+	keyConfigured  	             ContextKey = "arangodb-configured"
```
ewoutp: 1. The go-driver is vendored in this repo. Never change it here, without a PR in the go-driver repo first.
2. VST is controlled by the type of `Connection`, no need to add anything to the context for that.

pkg/util/arangod/client.go
```diff
@@ -156,20 +156,38 @@ func CreateArangodImageIDClient(ctx context.Context, deployment k8sutil.APIObjec
...
+		connConfig, err := createArangodVSTConfigForDNSNames(ctx, cli, apiObject, []string{dnsName})
+		if err != nil {
+			return nil, maskAny(err)
+		}
+		// TODO deal with TLS with proper CA checking
+		conn, err := vst.NewConnection(connConfig)
+		if err != nil {
+			return nil, maskAny(err)
+		}
+		// Create client
```
ewoutp: From this point on the code paths are the same. Only create `conn` in the if/else block.
Use `var conn driver.Connection` before the if/else block.
Then change `conn, err := ...` to `conn, err = ...`

pkg/util/arangod/client.go
```diff
@@ -156,20 +156,38 @@ func CreateArangodImageIDClient(ctx context.Context, deployment k8sutil.APIObjec
...
-	connConfig, err := createArangodHTTPConfigForDNSNames(ctx, cli, apiObject, []string{dnsName})
-	if err != nil {
-		return nil, maskAny(err)
-	}
-	// TODO deal with TLS with proper CA checking
-	conn, err := http.NewConnection(connConfig)
-	if err != nil {
-		return nil, maskAny(err)
+	config := driver.ClientConfig{}
+	vst := driver.mustUseVST(context)
```
ewoutp: I would not use the context for this, but add an argument to the function.
To avoid changing all uses of this function, use `vst ...bool` and then `if len(vst) > 0 && vst[0] { ...`

pkg/util/arangod/client.go
```diff
@@ -200,6 +218,22 @@ func createArangodHTTPConfigForDNSNames(ctx context.Context, cli corev1.CoreV1In
 	return connConfig, nil
 }
 
+// createArangodVSTConfigForDNSNames creates a go-driver VST connection config for a given DNS names.
+func createArangodVSTConfigForDNSNames(ctx context.Context, cli corev1.CoreV1Interface, apiObject *api.ArangoDeployment, dnsNames []string) (http.ConnectionConfig, error) {
+	scheme := "vst"
+	if apiObject != nil && apiObject.Spec.IsSecure() {
+		// TODO set security for transport
```
ewoutp: Secure connection is essential, since almost all kube-arangodb deployments use TLS.
To use a secure VST connection, set `scheme = "https"`.

pkg/util/arangod/client.go
```diff
@@ -200,6 +218,22 @@ func createArangodHTTPConfigForDNSNames(ctx context.Context, cli corev1.CoreV1In
 	return connConfig, nil
 }
 
+// createArangodVSTConfigForDNSNames creates a go-driver VST connection config for a given DNS names.
+func createArangodVSTConfigForDNSNames(ctx context.Context, cli corev1.CoreV1Interface, apiObject *api.ArangoDeployment, dnsNames []string) (http.ConnectionConfig, error) {
+	scheme := "vst"
```
ewoutp: Should be "http". There is no "vst" scheme in the go-driver, just a VST connection that re-uses http/https schemes.

pkg/util/arangod/client.go
```diff
@@ -186,7 +204,7 @@ func createArangodClientForDNSName(ctx context.Context, cli corev1.CoreV1Interfa
 func createArangodHTTPConfigForDNSNames(ctx context.Context, cli corev1.CoreV1Interface, apiObject *api.ArangoDeployment, dnsNames []string) (http.ConnectionConfig, error) {
 	scheme := "http"
 	transport := sharedHTTPTransport
-	if apiObject != nil && apiObject.Spec.IsSecure() {
+  if apiObject != nil && apiObject.Spec.IsSecure() {
```
ewoutp: Likely you did not use `go fmt`.

tests/load_balancer_test.go
```diff
@@ -0,0 +1,264 @@
...
+}
+
+type queryTestContext struct {
+	Context     context.Context
+	ExpectCount bool
+}
+
+
+// TestLoadBalancerCursorVST tests cursor forwarding with load-balanced VST
+func TestLoadBalancerCursorVST(t *testing.T) {
```
ewoutp: 1. Set the use of VST explicitly, not via context.
2. I would "duplicate" this test for VST & HTTP. To do so, create a helper function that contains almost all of the test and from `TestLoadBalancerCursorVST` and `TestLoadBalancerCursorHTTP` call that function with 1 argument specifying VST/HTTP.

tests/load_balancer_test.go
```diff
@@ -0,0 +1,264 @@
...
+
+	// Wait for deployment to be ready
+	apiObject, err := waitUntilDeployment(c, depl.GetName(), ns, deploymentIsReady())
+	if err != nil {
+		t.Fatalf("Deployment not running in time: %v", err)
+	}
+
+	// Create a database client
+	ctx := context.Background()
+	client := mustNewArangodDatabaseClient(ctx, kubecli, apiObject, t)
```
ewoutp: Extend this call to include VST/HTTP. (again as optional argument `vst ...bool`)

pkg/util/arangod/client.go
```diff
@@ -200,6 +218,22 @@ func createArangodHTTPConfigForDNSNames(ctx context.Context, cli corev1.CoreV1In
 	return connConfig, nil
 }
 
+// createArangodVSTConfigForDNSNames creates a go-driver VST connection config for a given DNS names.
+func createArangodVSTConfigForDNSNames(ctx context.Context, cli corev1.CoreV1Interface, apiObject *api.ArangoDeployment, dnsNames []string) (http.ConnectionConfig, error) {
+	scheme := "vst"
+	if apiObject != nil && apiObject.Spec.IsSecure() {
+		// TODO set security for transport
+	}
+	transport := vst.NewTransport(/* TODO configure transport */)
```
danielhlarkin: @ewoutp I haven't found examples of how to configure the transport correctly for secure vs. non-secure. Do you have a link to somewhere this is done?

tests/load_balancer_test.go
```diff
@@ -0,0 +1,273 @@
...
+			Query:             "FOR u IN users SORT u.name RETURN u.name",
+			ExpectedDocuments: []interface{}{"Blair", "Clair", "Jake", "John", "Johnny", "Zz"},
+			DocumentType:      reflect.TypeOf("foo"),
+			ExpectSuccess:     true,
+		},
+	}
+
+	// Setup context alternatives
+	contexts := []queryTestContext{
+		queryTestContext{driver.WithUseVST(nil, true), false},
```
ewoutp: remove WithUseVST

pkg/util/arangod/client.go
ewoutp: https://github.com/arangodb/go-driver/blob/master/vst/connection.go#L96

that line is what makes this work

pkg/util/arangod/client.go
```diff
@@ -65,23 +65,23 @@ var (
 		Proxy: nhttp.ProxyFromEnvironment,
 		DialContext: (&net.Dialer{
 			Timeout:   30 * time.Second,
-			KeepAlive: 30 * time.Second,
+			KeepAlive: 100 * time.Millisecond,
 			DualStack: true,
 		}).DialContext,
 		MaxIdleConns:          100,
-		IdleConnTimeout:       90 * time.Second,
+		IdleConnTimeout:       100 * time.Millisecond,
```
ewoutp: These are shared instances, so you're now changings this for ALL. That is not good.

tests/load_balancer_test.go
```diff
@@ -0,0 +1,280 @@
...
+	LoadBalancingCursorSubtest(t, false)
+}
+
+func wasForwarded(r *driver.Response) bool {
+	h := (*r).Header("x-arango-request-served-by")
+	return h != ""
+}
+
+// tests cursor forwarding with load-balanced conn.
+func LoadBalancingCursorSubtest(t *testing.T, useVst bool) {
```
ewoutp: This function should start with a lower-case letter.
Uppercase is used to export a member outside its package, which is not needed (and wanted) here.

tests/load_balancer_test.go
```diff
@@ -0,0 +1,280 @@
...
+	LoadBalancingCursorSubtest(t, true)
+}
+
+func TestLoadBalancingCursorHTTP(t *testing.T) {
+	// run with HTTP
+	LoadBalancingCursorSubtest(t, false)
+}
+
+func wasForwarded(r *driver.Response) bool {
+	h := (*r).Header("x-arango-request-served-by")
```
ewoutp: `driver.Response` is already an interface. Change `r *driver.Response` to `r driver.Response`.

tests/load_balancer_test.go
```diff
@@ -0,0 +1,280 @@
...
+	// Run tests for every context alternative
+	for _, qctx := range contexts {
+		ctx := qctx.Context
+		for i, test := range tests {
+			cursor, err := db.Query(ctx, test.Query, test.BindVars)
+			if err == nil {
+				// Close upon exit of the function
+				defer cursor.Close()
+			}
+			if test.ExpectSuccess {
```
ewoutp: This entire file needs `go fmt`. Since most editors do that automatically, any future change would cause a huge diff if you do not format now.

tests/test_util.go
```diff
@@ -60,6 +64,69 @@ var (
...
+		Transport: transport,
+	}
+	for _, dnsName := range dnsNames {
+		connConfig.Endpoints = append(connConfig.Endpoints, scheme+"://"+net.JoinHostPort(dnsName, strconv.Itoa(k8sutil.ArangoPort)))
+	}
+	return connConfig, nil
+}
+
+// CreateArangodDatabaseVSTClient creates a go-driver client for accessing the entire cluster (or single server) via VST
+func CreateArangodDatabaseVSTClient(ctx context.Context, cli corev1.CoreV1Interface, apiObject *api.ArangoDeployment) (driver.Client, error) {
```
ewoutp: start with lowercase letter. Should not be exported.

tests/load_balancer_test.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+func TestLoadBalancingCursorHTTP(t *testing.T) {
+	// run with HTTP
+	LoadBalancingCursorSubtest(t, false)
+}
+
+func wasForwarded(r *driver.Response) bool {
+	h := (*r).Header("x-arango-request-served-by")
+	return h != ""
```
ewoutp: API question. Is this header only set if the request is forwarded?
The header key seems to suggest that it is always there.
So I expected that you want to look for different values for this header instead of non-empty values.

tests/load_balancer_test.go
```diff
@@ -0,0 +1,280 @@
...
+						}
+						break
+					} else if err != nil {
+						t.Errorf("Failed to result document %d: %s", len(result), err)
+					}
+					if !hasMore {
+						t.Error("HasMore returned false, but ReadDocument returns a document")
+					}
+					result = append(result, doc.Elem().Interface())
+					if (wasForwarded(&r)) {
```
ewoutp: In go, if expressions do not need brackets. A `go fmt` will remove them automatically.

tests/load_balancer_test.go
```diff
@@ -0,0 +1,280 @@
...
+func LoadBalancingCursorSubtest(t *testing.T, useVst bool) {
+	c := client.MustNewInCluster()
+	kubecli := mustNewKubeClient(t)
+	ns := getNamespace(t)
+
+	// Prepare deployment config
+	depl := newDeployment("test-lb-" + uniuri.NewLen(4))
+	depl.Spec.Mode = api.NewMode(api.DeploymentModeCluster)
+	image := "dhly/arangodb:3.3.11-local"
+	depl.Spec.Image = &image
```
ewoutp: This works, but in order to be consistent with the rest of the tests, use:
`depl.Spec.Image = util.NewString("...")`

tests/load_balancer_test.go
danielhlarkin: Currently it is only set if the request is served by a different server than the one the request was sent to. You are correct that this is less than intuitive, so I'll think a bit about a better name.

PR #227, opened 2018-08-02
indicate that definitions aren't versioned alone

janat08:


janat08:
Apparently you can do this electronically as in: https://contribute.meteor.com/

Issue #251, opened 2018-09-07
Finalizers preventing namespace termination


rglyons:
I'm trying to delete a namespace in its entirety, however the arangodb resource finalizers are preventing that.

Say I have a namespace called `example`. Now say there is an ArangoDeployment running in this namespace. When I run `kubectl delete ns example`, The ArangoDeployment is triggered to delete. However, the finalizers on the ArangoDeployment are never removed, since the namespace deletion has already terminated the arangodb operator, because the operator has no finalizers. This means that  the ArangoDeployment is never deleted. Also, the arangodb PVCs are never deleted, since they have finalizers that wait for the ArangoDeployment to tear down before they fully terminate. All in all, the namespace `example` is stuck terminating forever. 

How can I safely tear down a namespace where an ArangoDeployment and arangodb operator are running? The only option I see is to delete the ArangoDeployment first, then wait for it to delete (along with its PVCs), then _finally_ execute `kubectl delete ns example`. The problem I have with this workflow is that it requires scripting that would disrupt many of my existing workflows. 

One idea that I had was to add finalizers to the arangodb operator that prevent it from fully terminating _until_ all arangodb resources have fully terminated. This, however, would take away the ability to deploy arangodb on k8s using the operator, then delete the operator, which is a use case some might have, assuming they never want to update/scale arangodb on their cluster. Whether or not you find this use case valid as something you want to support determines the validity of this potential solution.

If not the above proposed solution, what other solutions are there?

Arabus:
Ran into the same issue twice now - only solution I found was to kill the whole cluster. I'd be grateful for a workaround when one has already deleted the operator. Is it possible to manually remove the finalizers?

maierlars:
Yes you can just remove the finalizers.

Arabus:
Apparently my question was a little misunderstandable so here is the answer I expected:

```
kubectl get ArangoDeployment <ArangoDeployment-instance-name> -o=json | \
jq '.metadata.finalizers = null' | kubectl apply -f -
```

This works the same with all other objects that have finalizers in them relying on the operator.

For a more permanent solution I recommend deploying the operator to a different namespace (You might need to adjust the rbac rules).

maxzzze:
Reviving an old thread here but its the only one I can find. Ran into a similar issue. Tried draining a node and its stuck because of the arango prmr pods stuck terminating. 

I tried the solution above to remove finalizers and also created a new operator on a new node but the pods are still stuck.

[...]

PR #253, opened 2018-09-12
added a flag in the ArangoDeployment K8s spec to turn off finalizers …

rglyons:
…in ArangoDeployments and their PVCs

The default behavior is still to apply, validate, and respect finalizers on the ArangoDeployment resources, however this behavior can be overwritten with the `enableFinalizers` boolean value in the ArangoDeployment spec. In the case that `enableFinalizers: false` is provided in the ArangoDeployment yaml spec, all agents, coordinators, db servers, and pvcs are brought up with no finalizers. This allows users to teardown entire clusters in any order, and was implemented to serve the use case of single-step (unordered) namespace teardowns in kubernetes. 

Please let me know if I have missed any steps (writing/running some test suite, etc.). I am happy to work with you to get this merged, because maintaining a custom fork of this repo could be cumbersome. 

related to #251 

NOTE: if this PR is accepted and merged, this documentation page will need to be updated - 
https://docs.arangodb.com/devel/Manual/Deployment/Kubernetes/DeploymentResource.html#specsyncexternalaccesstype-string


OmarAyo:
Hi @rglyons 

Thanks a lot for your improvements.  In order to merge it, we need a signed CLA.  We are using an Apache 2 CLA for ArangoDB and its sub-components, which can be found here: https://www.arangodb.com/documents/cla.pdf

Could you please sign, scan and email it to cla@arangodb.com

Thanks a lot

Omar Ayoubi

ewoutp:
hi @rglyons 
While I understand the intent of this PR I have serious concerns about.
Without the finalizers, there is no way we can guarantee data safety.

We'll discuss what we want to do with this and then come back on it.

rglyons:
@ewoutp I understand your concerns and why your design includes finalizers. This PR does not change the default behavior - finalizers are still applied _unless_ the `enableFinalizers: false` option is provided in the ArangoDeployment yaml. This gives the user the option to forego that data protection _if they choose to do so_. 

Ideally, the finalizers remain, but full teardown is controlled so that the system never ends up in a frozen state, e.g. when the operator completes teardown before any ArangoDB resources. If that were implemented then this PR would be obsolete. 

I'm looking forward to hearing about the conclusions from your discussions.

neunhoef:
Hi @rglyons,
after some discussions we have decided against incorporating this PR, sorry for this.

Here is our thinking:

This would be a switch which set to `false` would seriously harm the resilience of the system, since deployed resources could be deleted without warning, which can very easily result in data loss. This is the worst that can ever happen to a database. We think that once this option is available, we will eventually have people who have bought support and still use this flag. This can lead to extremely bad situations, which we have to avoid at all cost. And even if we find this usage before it is too late, with this PR we cannot even change the deployment easily after the fact, since one cannot edit the flag after deployment. Implementing this functionality would be an undue burden for development and unduly complicate things.

We understand that this has the potential to cause inconveniences during development, but we think this can be avoided easily by removing all custom resources before removing the operator deployment itself. So in the end, this is a judgement weighing potential business risk against developer convenience.

If you have a scenario which we have overlooked in this reasoning and which would create a huge inconvenience, please let us know. I am closing this for now, feel free to open again if there are new aspects.

Cheers,
  Max.

PR #262, opened 2018-10-05
Maintenance mode for upgrades

graetzer:
Enable Maintenance Mode when updating an ActiveFailover deployment.


lamai93:
http://c11:8080/job/kube-arangodb-branches/job/feature%252Fmaintenance-mode/5/

PR #275, opened 2018-10-25
Bug fix/scale up error

lamai93:
Do not update the agency if not required. This could reset some user input.


pkg/deployment/reconcile/plan_builder.go
```diff
@@ -366,13 +366,22 @@ func createScalePlan(log zerolog.Logger, members api.MemberStatusList, group api
...
+					)
+				}
 				plan = append(plan,
-					api.NewAction(api.ActionTypeCleanOutMember, group, m.ID),
+					api.NewAction(api.ActionTypeShutdownMember, group, m.ID),
 				)
 			}
 			plan = append(plan,
-				api.NewAction(api.ActionTypeShutdownMember, group, m.ID),
 				api.NewAction(api.ActionTypeRemoveMember, group, m.ID),
```
lamai93: Put this logic into plan execution and keep the plan as it was. Reason: reduce time one can screw things up by killing a running dbserver.

PR #276, opened 2018-10-25
Fixed readiness route.

lamai93:


pkg/deployment/resources/pod_creator.go
```diff
@@ -369,7 +369,7 @@ func (r *Resources) createReadinessProbe(spec api.DeploymentSpec, group api.Serv
 		}
 	}
 	probeCfg := &k8sutil.HTTPProbeConfig{
-		LocalPath:           "/_api/version",
+		LocalPath:           "/_admin/server/availability",
```
lamai93: Make this more robust and choose route depending on version.

PR #294, opened 2018-11-05
Fixing uninitialised `lastNumberOfServers`.

lamai93:
Previously the spec was updated every two seconds with the values written in the agency. This was due to a change that was done so that the operator will respect the ui after being redeployed.

The root cause seems to be a uninitialised lastNumberOfServers struct.

Ignore debug output.

### DIFF
--- a/pkg/deployment/cluster_scaling_integration.go
+++ b/pkg/deployment/cluster_scaling_integration.go
@@ -65,6 +65,7 @@ func newClusterScalingIntegration(depl *Deployment) *clusterScalingIntegration {
 
 // SendUpdateToCluster records the given spec to be sended to the cluster.
 func (ci *clusterScalingIntegration) SendUpdateToCluster(spec api.DeploymentSpec) {
+	ci.log.Debug().Msg("SendUpdateToCluster called")
 	ci.pendingUpdate.mutex.Lock()
 	defer ci.pendingUpdate.mutex.Unlock()
 	ci.pendingUpdate.spec = &spec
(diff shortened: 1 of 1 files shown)


pkg/deployment/cluster_scaling_integration.go
```diff
@@ -132,15 +138,32 @@ func (ci *clusterScalingIntegration) inspectCluster(ctx context.Context, expectS
...
 		// #Coordinator has changed
 		coordinatorsChanged = true
 	}
-	if req.DBServers != nil && req.GetDBServers() != desired.GetDBServers() {
+	if req.DBServers != nil && desired.DBServers != nil && req.GetDBServers() != desired.GetDBServers() {
 		// #DBServers has changed
 		dbserversChanged = true
 	}
 	if !coordinatorsChanged && !dbserversChanged {
+		// if there is nothing to change, check if we naver have asked the cluster before
```
neunhoef: typo naver

pkg/deployment/cluster_scaling_integration.go
```diff
@@ -132,15 +138,32 @@ func (ci *clusterScalingIntegration) inspectCluster(ctx context.Context, expectS
...
+		// if so, fill in the values for the first time.
+		// This happens, when the operator is redeployed and there has not been any
+		// update events yet.
+		if desired.Coordinators == nil || desired.DBServers == nil {
+			//ci.lastNumberOfServers.mutex.Lock()
```
neunhoef: Why is this commented out, it looks right!

lamai93: because its already locked 15 line above

PR #318, opened 2019-01-01
Use jwt-keyfile option if available.

lamai93:
This is untested because no image available.


pkg/deployment/resources/pod_creator.go
```diff
@@ -85,8 +96,17 @@ func createArangodArgs(apiObject metav1.Object, deplSpec api.DeploymentSpec, gro
 		// With authentication
 		options = append(options,
 			optionPair{"--server.authentication", "true"},
-			optionPair{"--server.jwt-secret", "$(" + constants.EnvArangodJWTSecret + ")"},
 		)
+		if versionHasJWTSecretKeyfile(version) {
+			keyPath := filepath.Join(k8sutil.ClusterJWTSecretVolumeName, constants.SecretKeyToken)
```
ewoutp: Don't think using the volume name is correct.
Should be the mount path of that volume (`ClusterJWTSecretVolumeMountDir`)

pkg/deployment/resources/pod_creator.go
```diff
@@ -65,6 +65,17 @@ func versionHasAdvertisedEndpoint(v driver.Version) bool {
 	return v.CompareTo("3.4.0") >= 0
 }
 
+func versionHasJWTSecretKeyfile(v driver.Version) bool {
```
ewoutp: Please add comment

pkg/deployment/resources/pod_creator.go
neunhoef: Right.

pkg/util/k8sutil/pods.go
```diff
@@ -447,6 +448,9 @@ func CreateArangodPod(kubecli kubernetes.Interface, developmentMode bool, deploy
 	if rocksdbEncryptionSecretName != "" {
 		c.VolumeMounts = append(c.VolumeMounts, rocksdbEncryptionVolumeMounts()...)
 	}
+	if ClusterJWTSecretVolumeName != "" {
```
neunhoef: I think this will always be true.

neunhoef: Should be `clusterJWTSecretName`

pkg/deployment/resources/pod_creator.go
neunhoef: Done.

pkg/util/k8sutil/pods.go
lamai93: Yes

PR #321, opened 2019-01-07
Added scaling limits.

lamai93:


pkg/apis/deployment/v1alpha/server_group_spec.go
```diff
@@ -110,8 +114,27 @@ func (s ServerGroupSpec) Validate(group ServerGroup, used bool, mode DeploymentM
 				minCount = 2
 			}
 		}
+		var specMinCount int
```
ewoutp: Suggest to add helper functions `GetMinCount() int` & `GetMaxCount() int` that return the effective min/max count values.
If the setting is nil, return return 1/maxInt.

Then in Validate use those functions to avoid nil checks in there.

pkg/apis/deployment/v1alpha/server_group_spec_test.go
```diff
@@ -48,6 +48,11 @@ func TestServerGroupSpecValidateCount(t *testing.T) {
 	assert.Nil(t, ServerGroupSpec{Count: util.NewInt(2)}.Validate(ServerGroupSyncMasters, true, DeploymentModeCluster, EnvironmentProduction))
 	assert.Nil(t, ServerGroupSpec{Count: util.NewInt(2)}.Validate(ServerGroupSyncWorkers, true, DeploymentModeCluster, EnvironmentProduction))
 
+	assert.Nil(t, ServerGroupSpec{Count: util.NewInt(2), MinCount: util.NewInt(2), MaxCount: util.NewInt(5)}.Validate(ServerGroupCoordinators, true, DeploymentModeCluster, EnvironmentDevelopment))
```
ewoutp: You do test the edge for minCount (count=2, minCount=2) but not for maxCount.
Suggest to add count=5,minCount=5,maxCount=5

pkg/apis/deployment/v1alpha/server_group_spec.go
```diff
@@ -58,6 +63,22 @@ func (s ServerGroupSpec) GetCount() int {
 	return util.IntOrDefault(s.Count)
 }
 
+// GetMinCount returns MinCount or 1 if not set
+func (s ServerGroupSpec) GetMinCount() int {
+	if s.MinCount != nil {
```
ewoutp: Use `return util.IntOrDefault(s.MinCount, 1)`

docs/Manual/Deployment/Kubernetes/DeploymentResource.md
```diff
@@ -352,6 +352,14 @@ for `spec.mode: Single` and `2` for `spec.mode: ActiveFailover`).
 For the `syncworkers` group, it is highly recommended to use the same number
 as for the `dbservers` group.
 
+### `spec.<group>.minCount: number`
+
+Specifies a minimum for the count of servers. If set, a specification is invalid if `count < minCount`.
```
ewoutp: What about `minCount <= maxCount`?

pkg/apis/deployment/v1alpha/server_group_spec.go
```diff
@@ -110,8 +131,14 @@ func (s ServerGroupSpec) Validate(group ServerGroup, used bool, mode DeploymentM
 				minCount = 2
 			}
 		}
+		if s.GetCount() < s.GetMinCount() {
```
ewoutp: Forgot earlier, check minCount <= maxCount

pkg/apis/deployment/v1alpha/server_group_spec.go
lamai93: Sure.

docs/Manual/Deployment/Kubernetes/DeploymentResource.md
lamai93: This is implicit, since there is no `count` that could possible satisfy both: 
if there is `c` with `a <= c` and `c <= b` then `a <= b`.

pkg/apis/deployment/v1alpha/server_group_spec.go
lamai93: Yeah I can add that check.

PR #325, opened 2019-01-14
Randomize Root Password

maierlars:
Randomize root password or use provided secret.
This is a breaking change.


neunhoef:
There are still open questions:
(1) This is a breaking change, since all documentation is wrong now.
(2) What happens in an upgrade case if the operator is upgraded?
(3) How should the field in the spec be called and what is its semantics?

neunhoef:
Documentation missing.

pkg/apis/deployment/v1alpha/deployment_spec.go
```diff
@@ -130,6 +133,10 @@ func (s DeploymentSpec) IsSecure() bool {
 	return s.TLS.IsSecure()
 }
 
+func (s DeploymentSpec) GetRootUserAccessSecretName() string {
```
ewoutp: Comment missing

pkg/apis/deployment/v1alpha/deployment_spec.go
```diff
@@ -70,6 +71,8 @@ type DeploymentSpec struct {
 	SyncWorkers  ServerGroupSpec `json:"syncworkers"`
 
 	Chaos ChaosSpec `json:"chaos"`
+
+	RootUserAccessSecretName *string `json:"rootUserAccessSecretName,omitempty"`
```
ewoutp: Don't understand the name.
Why not `RootUserPasswordSecretName`?

pkg/apis/deployment/v1alpha/deployment_spec.go
```diff
@@ -277,6 +290,9 @@ func (s *DeploymentSpec) Validate() error {
 	if err := s.License.Validate(); err != nil {
 		return maskAny(errors.Wrap(err, "spec.licenseKey"))
 	}
+	if err := k8sutil.ValidateResourceName(s.GetRootUserAccessSecretName()); err != nil {
```
ewoutp: I expect that I can also set it to None.

pkg/deployment/bootstrap.go
```diff
@@ -0,0 +1,129 @@
...
+			return maskAny(err)
+		}
+
+		d.deps.Log.Info().Msgf("Bootstrap completed for %s", d.Name())
+	}
+
+	return nil
+}
+
+func (d *Deployment) ensureRootUserPassword() (string, error) {
```
ewoutp: comments?

pkg/deployment/bootstrap.go
```diff
@@ -0,0 +1,129 @@
...
+				return string(pass), nil
+			}
+		}
+		return "", fmt.Errorf("invalid secret format")
+	} else {
+		return "", err
+	}
+}
+
+func (d *Deployment) runBootstrap() error {
```
ewoutp: comments...

pkg/deployment/bootstrap.go
```diff
@@ -0,0 +1,129 @@
...
+		return maskAny(err)
+	}
+
+	// Obtain the root user
+	root, err := client.User(nil, rootUserName)
+	if err != nil {
+		return maskAny(err)
+	}
+
+	err = root.Update(nil, driver.UserOptions{
```
ewoutp: For statements like this, I'll prefer `if err := root.Update...(); err != nil {....}` to narrow the scope as much as possible

pkg/deployment/bootstrap.go
```diff
@@ -0,0 +1,129 @@
...
+func (d *Deployment) EnsureBootstrap() error {
+
+	status, version := d.GetStatus()
+
+	if status.Conditions.IsTrue(api.ConditionTypeReady) {
+		if status.Conditions.IsTrue(api.ConditionTypeBoostrapCompleted) {
+			return nil // Nothing to do, already bootstrapped
+		}
+
+		d.deps.Log.Info().Msgf("Bootstrap deployment %s", d.Name())
```
ewoutp: This will likely run once for existing deployments.

That sounds like a problem.

pkg/apis/deployment/v1alpha/deployment_spec.go
ewoutp: Explicitly setting this to `None` means that you do not want the operator to set any root password.
Therefore this should NOT be the default.

pkg/apis/deployment/v1alpha/bootstrap.go
```diff
@@ -29,46 +29,66 @@ const (
...
 
 // BootstrapSpec contains information for cluster bootstrapping
 type BootstrapSpec struct {
 	// PasswordSecretNames contains a map of username to password-secret-name
 	PasswordSecretNames PasswordSecretNameList `json:"passwordSecretNames,omitempty"`
 }
 
+// IsNone returns ture if p is empty or None
+func (p PasswordSecretName) IsNone() bool {
+	return p == PasswordSecretNameNone || p == ""
```
ewoutp: Isn't an empty string supposed to be ""Auto"?

pkg/apis/deployment/v1alpha/bootstrap.go
```diff
@@ -78,13 +98,13 @@ func (b *BootstrapSpec) Validate() error {
 // SetDefaults fills in default values when a field is not specified.
 func (b *BootstrapSpec) SetDefaults(deploymentname string) {
```
ewoutp: This function takes me long to understand.
You could simply by having these 3 steps:
- check if map is nil -> make new (empty) map
- check if map contains root, if not add it with Auto value
- foreach entry in map, if Auto -> replace with calculated secret name

pkg/apis/deployment/v1alpha/bootstrap.go
```diff
@@ -29,46 +29,66 @@ const (
...
 
 // Validate the specification.
 func (b *BootstrapSpec) Validate() error {
 	for _, secretname := range b.PasswordSecretNames {
-		if secretname == PasswordSecretNameNone {
-			continue
-		}
-		if err := k8sutil.ValidateResourceName(secretname); err != nil {
-			return maskAny(err)
+		if err := secretname.Validate(); err != nil {
```
ewoutp: Should None be allowed for anything other than the `root` user?
Don't think it makes a lot of sense for others.

pkg/apis/deployment/v1alpha/bootstrap.go
```diff
@@ -0,0 +1,143 @@
...
+// PasswordSecretNameList is a map from username to secretnames
+type PasswordSecretNameList map[string]PasswordSecretName
+
+// BootstrapSpec contains information for cluster bootstrapping
+type BootstrapSpec struct {
+	// PasswordSecretNames contains a map of username to password-secret-name
+	PasswordSecretNames PasswordSecretNameList `json:"passwordSecretNames,omitempty"`
+}
+
+// IsNone returns ture if p is empty or None
```
ewoutp: ture -> true (twice)

ewoutp: And now code is correct, but does not match comment

pkg/deployment/bootstrap.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+
+		if status.Conditions.IsTrue(api.ConditionTypeBoostrapCompleted) {
+			return nil // Nothing to do, already bootstrapped
+		}
+
+		d.deps.Log.Info().Msgf("Bootstrap deployment %s", d.Name())
+		err := d.runBootstrap()
+		if err != nil {
+			status.Conditions.Update(api.ConditionTypeBoostrapCompleted, true, "Bootstrap failed", err.Error())
```
ewoutp: Why not add a `BootstrapSucceeded` condition?
No I have to check for a string, which may change

PR #328, opened 2019-01-17
Updated deps. Multiarch build.

maierlars:


Makefile
```diff
@@ -212,35 +225,44 @@ dashboard/assets.go: $(DASHBOARDSOURCES) $(DASHBOARDDIR)/Dockerfile.build
...
 .PHONY: docker
-docker: check-vars $(BIN)
-	docker build -f $(DOCKERFILE) -t $(OPERATORIMAGE) .
-ifdef PUSHIMAGES
-	docker push $(OPERATORIMAGE)
-endif
+docker: check-vars build-bin $(MANIFESTTOOL)
+	for arch in $(ARCHS); do \
+		docker build --build-arg=GOARCH=$$arch -t $(OPERATORIMAGE)-$$arch . ;\
+		docker push $(OPERATORIMAGE)-$$arch ;\
```
ewoutp: Think you need to wrap this in `ifdef PUSHIMAGES` ... `endif`

Makefile
```diff
@@ -212,35 +225,44 @@ dashboard/assets.go: $(DASHBOARDSOURCES) $(DASHBOARDDIR)/Dockerfile.build
...
+	done
+	$(MANIFESTTOOL) $(MANIFESTAUTH) push from-args \
```
ewoutp: Think you need to wrap this in `ifdef PUSHIMAGES` ... `endif`

ewoutp: This does not run (out of the box) on macos.
The problem is docker registry authentication. That's why `MANIFESTAUTH` environment variable is in there.
It has to be set to something like `--username=foo --password=bar`.

See https://github.com/estesp/manifest-tool#sample-usage

Suggest to add this to the README in the building section.

PR #335, opened 2019-01-30
Clean out action

lamai93:
Do not trigger a new clean out action when the server is already cleaned out.

### DIFF
--- a/pkg/deployment/reconcile/action_cleanout_member.go
+++ b/pkg/deployment/reconcile/action_cleanout_member.go
@@ -77,6 +77,13 @@ func (a *actionCleanoutMember) Start(ctx context.Context) (bool, error) {
 	}
 	if cleanedOut {
 		log.Debug().Msg("Server already cleaned out")
+		// Cleanout completed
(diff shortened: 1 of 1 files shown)


pkg/deployment/reconcile/action_cleanout_member.go
```diff
@@ -70,6 +70,15 @@ func (a *actionCleanoutMember) Start(ctx context.Context) (bool, error) {
 		log.Debug().Err(err).Msg("Failed to access cluster")
 		return false, maskAny(err)
 	}
+	cleanedOut, err := cluster.IsCleanedOut(ctx, a.action.MemberID)
+	if err != nil {
+		log.Debug().Err(err).Msg("IsCleanedOut failed")
+		return false, maskAny(err)
+	}
+	if cleanedOut {
+		log.Debug().Msg("Server already cleaned out")
```
ewoutp: Should you set the `ConditionTypeCleanedOut` condition here, see end of `CheckProgress`

PR #338, opened 2019-02-11
Fix Reload

maierlars:
Added secret force-status-reload status field. ✌️

### DIFF
--- a/pkg/apis/deployment/v1alpha/deployment_status.go
+++ b/pkg/apis/deployment/v1alpha/deployment_status.go
@@ -59,7 +59,7 @@ type DeploymentStatus struct {
 	// detect changes in secret values.
 	SecretHashes *SecretHashes `json:"secret-hashes,omitempty"`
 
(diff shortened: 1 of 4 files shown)


pkg/apis/deployment/v1alpha/deployment_status.go
```diff
@@ -56,6 +58,9 @@ type DeploymentStatus struct {
 	// SecretHashes keeps a sha256 hash of secret values, so we can
 	// detect changes in secret values.
 	SecretHashes *SecretHashes `json:"secret-hashes,omitempty"`
+
+	// ForceStatusReload if set to true forces a reload of the status from the CR.
```
ewoutp: CR?

lamai93: CustomResource

lamai93:
This would allow easy patching of the status, when the operator is not able to recover by itself - without removing and redeploying the operator which might not be that easy.

PR #339, opened 2019-02-11
No Coords

maierlars:
Delete all coordinator pods if all coordinators are failed. If a pod is in strange phase but there is no plan, view it as failed.

### DIFF
--- a/pkg/apis/deployment/v1alpha/member_status_list.go
+++ b/pkg/apis/deployment/v1alpha/member_status_list.go
@@ -183,25 +183,24 @@ func (l MemberStatusList) AllMembersReady() bool {
 
 // AllConditionTrueSince returns true if all members satisfy the condition since the given period
 func (l MemberStatusList) AllConditionTrueSince(cond ConditionType, status v1.ConditionStatus, period time.Duration) bool {
-	trueCount := 0
 	for _, x := range l {
 		if c, ok := x.Conditions.Get(cond); ok {
 			if c.Status == status && c.LastTransitionTime.Time.Add(period).Before(time.Now()) {
-				trueCount++
+				continue
 			}
 		}
+		return false
 	}
 
-	return trueCount == len(l)
+	return true
 }
 
 // AllFailed returns true if all members are failed
 func (l MemberStatusList) AllFailed() bool {
-	failedCount := 0
 	for _, x := range l {
-		if x.Phase.IsFailed() {
-			failedCount++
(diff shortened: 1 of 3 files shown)


pkg/apis/deployment/v1alpha/member_status_list.go
```diff
@@ -178,3 +180,28 @@ func (l MemberStatusList) MembersReady() int {
...
 	return len(l) == l.MembersReady()
 }
+
+// AllConditionTrueSince returns true if all members satisfy the condition since the given period
+func (l MemberStatusList) AllConditionTrueSince(cond ConditionType, status v1.ConditionStatus, period time.Duration) bool {
+	trueCount := 0
+	for _, x := range l {
+		if c, ok := x.Conditions.Get(cond); ok {
+			if c.Status == status && c.LastTransitionTime.Time.Add(period).Before(time.Now()) {
+				trueCount++
```
ewoutp: Why not simply return `false` if condition is not met?

pkg/apis/deployment/v1alpha/member_status_list.go
```diff
@@ -178,3 +180,28 @@ func (l MemberStatusList) MembersReady() int {
...
+
+	return trueCount == len(l)
+}
+
+// AllFailed returns true if all members are failed
+func (l MemberStatusList) AllFailed() bool {
+	failedCount := 0
+	for _, x := range l {
+		if x.Phase.IsFailed() {
+			failedCount++
```
ewoutp: Why not simply return `false` if condition is not met?

pkg/deployment/resources/pod_inspector.go
```diff
@@ -237,6 +237,19 @@ func (r *Resources) InspectPods(ctx context.Context) (util.Interval, error) {
 	allMembersReady := status.Members.AllMembersReady(spec.GetMode(), spec.Sync.IsEnabled())
 	status.Conditions.Update(api.ConditionTypeReady, allMembersReady, "", "")
 
+	if status.Members.Coordinators.AllFailed() {
```
ewoutp: This expression seems to "soon".
Shouldn't it given the coordinators some time first?

lamai93: If they are failed, the operator will create a member replacement plan anyways. However, if all are failed, the execution of the plan will always fail because it can not connect to the cluster. So no, if all coordinators are failed, there is nothing we can do.

PR #343, opened 2019-02-21
Add instructions for k8s node drain operation.

neunhoef:


docs/Manual/Deployment/Kubernetes/Drain.md
```diff
@@ -0,0 +1,444 @@
...
+takes time, and even after the move, the distributed system ArangoDB has
+to recover from this change, for example by ensuring data synchronicity
+between the replicas in their new location.
+
+{% hint 'warning' %}
+Therefore, a systematic drain of all k8s nodes in sequence has to follow
+a careful procedure, in particular to ensure that ArangoDB is ready to
+move to the next step. This is necessary to avoid catastrophic data
+loss, and is simply the price one pays for running a stateful service.
+{% endhint %}
```
Simran-B: @sleto-it I added a red hint box at the top, so maybe we can remove your warning here?

docs/Manual/Deployment/Kubernetes/Drain.md
```diff
@@ -0,0 +1,444 @@
...
+  "code": 200
+}
+```
+
+Check that each instance has a `Status` field with the value `"GOOD"`.
+Here is a shell command which makes this check easy, using the
+[`jq` JSON pretty printer](https://stedolan.github.io/jq/):
+
+```bash
+curl -k https://arangodb.9hoeffer.de:8529/_admin/cluster/health --user root: | jq . | grep '"Status"' | grep -v '"GOOD"'
```
Simran-B: @neunhoef Does jq support something similar to grep, so that we can make this a cross-platform command?

docs/Manual/Deployment/Kubernetes/Drain.md
```diff
@@ -0,0 +1,444 @@
...
+}
+```
+
+Check that for all collections the attribute `"allInSync"` has
+the value `true`. Note that it is necessary to do this for all databases!
+
+Here is a shell command which makes this check easy:
+
+```bash
+curl -k https://arangodb.9hoeffer.de:8529/_db/_system/_api/replication/clusterInventory --user root: | jq . | grep '"allInSync"' | sort | uniq -c
```
Simran-B: Does jq support this with a filter? There is group_by()...

docs/Manual/Deployment/Kubernetes/Drain.md
```diff
@@ -0,0 +1,444 @@
...
+POST /_admin/cluster/cleanOutServer
+```
+
+… and send as body a JSON document like this:
+
+```JSON
+{"server":"DBServer0006"}
+```
+
+(please compare the above output of the `/_admin/cluster/health` API).
```
Simran-B: @neunhoef  compare the cleanOutServer output to the cluster health output?

docs/Manual/Deployment/Kubernetes/Drain.md
```diff
@@ -0,0 +1,444 @@
...
+The value of the `"server"` attribute should be the name of the DBserver
+which is one the pod which shall be drained next. This uses the UI short
```
Simran-B: @neunhoef like this?
```suggestion
which is one of the pods which shall be drained next. This uses the UI short
```

docs/Manual/Deployment/Kubernetes/Drain.md
```diff
@@ -0,0 +1,444 @@
...
+name, alternatively one can use the internal name, which corresponds to
+the pod name: In our example, the pod name is:
```
Simran-B:
```suggestion
the pod name. In our example, the pod name is:
```

docs/Manual/Deployment/Kubernetes/Drain.md
```diff
@@ -0,0 +1,444 @@
...
+of data can take considerably longer!
+
+If the optional step of
+[cleaning out a DBserver manually](#clean-out-a-dbserver-manually-optional)
+has been performed beforehand, the grace period can easily be reduced to 60
+seconds - at least from the perspective of ArangoDB, since the server is already
+cleaned out, so it can be dropped readily and there is still no risk.
+
+At the same time, this guarantees now that the drain is completed
+approximately within a minute.
```
Simran-B: Sounds like cleaning out the server manually is what we should recommend, yet we say it's optional. What are the other options? The problem I see with any grace period is that you never know if it's too long or too short, and that the draining process never finish before the end of the grace period...

docs/Manual/Deployment/Kubernetes/Drain.md
```diff
@@ -0,0 +1,444 @@
...
+
+## Things to check after a node drain
+
+After a node has been drained, there will usually be one of the
+_DBservers_ gone from the cluster. As a replacement, another _DBServer_ has
+been deployed on a different node, if there is a different node
+available. If not, the replacement can only be deployed when the
+maintenance work on the drained node has been completed and it is
+uncordoned again. In this latter case, one should wait until the node is
```
Simran-B: is `uncordoned` a real word?

docs/Manual/Deployment/Kubernetes/Drain.md
```diff
@@ -0,0 +1,444 @@
...
+back up and the replacement pod has been deployed there.
+
+After that, one should perform the same checks as described in Section
```
Simran-B:
```suggestion
After that, one should perform the same checks as described in
```

docs/Manual/Deployment/Kubernetes/Drain.md
```diff
@@ -0,0 +1,444 @@
...
+
+After that, one should perform the same checks as described in Section 
+[things to check before a node drain](#things-to-check-in-arangodb-before-a-node-drain)
+above.
+
+Finally, it is likely that the shard distribution in the "new" cluster
+is not balanced out. In particular, the new _DBSserver_ is not automatically
+used to store shards. We recommend to
+[re-balance](../../Administration/Cluster/#movingrebalancing-shards) the shard distribution,
+either manually by moving shards or by using the "Rebalance Shards"
```
Simran-B:
```suggestion
either manually by moving shards or by using the _Rebalance Shards_
```

docs/Manual/Deployment/Kubernetes/Drain.md
```diff
@@ -0,0 +1,444 @@
...
+button in the "NODES/Shards" tab in the UI. This redistribution can take
```
Simran-B:
```suggestion
button in the _Shards_ tab under _NODES_ in the web UI. This redistribution can take
```

sleto-it:
all the open conversation will be considered for a future pr

lservini:
LGTM

PR #352, opened 2019-03-07
Added pod disruption budgets in production mode.

maierlars:


pkg/deployment/deployment.go
```diff
@@ -215,6 +215,11 @@ func (d *Deployment) run() {
 			d.CreateEvent(k8sutil.NewErrorEvent("Failed to create pods", err, d.GetAPIObject()))
 		}
 
+		// Create PDBs
```
ewoutp: Probably helpful to spell our PodDisruptionBudgets completely in this comment. PDB is not that well known

pkg/deployment/resources/pdbs.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// EnsurePDBs ensures Pod Distruption Bugets for different server groups in Cluster mode
+func (r *Resources) EnsurePDBs() error {
+
+	// Only in Cluster and Production Mode
+	spec := r.context.GetSpec()
+	if spec.IsProduction() && spec.GetMode().IsCluster() {
+
+		minAgents := spec.GetServerGroupSpec(api.ServerGroupAgents).GetCount() - 1
```
ewoutp: Please add comment to explain these rules.
This is probably the key part of the entire PR

pkg/deployment/resources/pdbs.go
```diff
@@ -0,0 +1,105 @@
...
+
+		if err := r.ensurePDBForGroup(api.ServerGroupAgents, minAgents); err != nil {
+			return err
+		}
+		if err := r.ensurePDBForGroup(api.ServerGroupDBServers, minDBServers); err != nil {
+			return err
+		}
+		if err := r.ensurePDBForGroup(api.ServerGroupCoordinators, minCoordinators); err != nil {
+			return err
+		}
```
ewoutp: What about sync masters & sync workers?

pkg/deployment/resources/pdbs.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	return nil
+}
+
+func pdbNameForGroup(depl string, group api.ServerGroup) string {
+	return fmt.Sprintf("%s-%s-pdb", depl, group.AsRole())
```
ewoutp: Will these names not exceed k8s name limits?

pkg/deployment/resources/pdbs.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func newPDB(minAvail int, deplname string, group api.ServerGroup, owner metav1.OwnerReference) *policyv1beta1.PodDisruptionBudget {
+	return &policyv1beta1.PodDisruptionBudget{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            pdbNameForGroup(deplname, group),
+			OwnerReferences: []metav1.OwnerReference{owner},
```
ewoutp: What about namespace?

pkg/deployment/resources/pdbs.go
```diff
@@ -0,0 +1,105 @@
...
+			}
+			return nil
+		}
+
+	}
+
+	return maskAny(err)
+}
+
+func newFromInt(v int) *intstr.IntOrString {
```
ewoutp: Replace (or use) with `intstr.FromInt(...)` ?

maierlars: `intstr.FromInt` returns by value, I need by pointer. I could do 
```
ret = &intstr.IntOrString
*ret = intstr.FromInt(v)
```

pkg/deployment/resources/pdbs.go
maierlars: Namespace is implicitly set due by the api-server, since the client-set already has the namespace.

pkg/deployment/resources/pdbs.go
maierlars: I think this is a general problem with default name, too. For example ` deploymentName+"-sync-client-auth-ca"`. Or all those TLS certificates. As the upper limit is 253, you have ~240 characters for you deployment name. I think this is not a problem.

ewoutp: Some kubernetes names can be longer than others.
I've seen problems with it, so at least be aware of it.

pkg/deployment/deployment.go
```diff
@@ -215,6 +215,11 @@ func (d *Deployment) run() {
 			d.CreateEvent(k8sutil.NewErrorEvent("Failed to create pods", err, d.GetAPIObject()))
 		}
 
+		// Create Pod Distruption Budgets
```
ewoutp: Distruption -> Disruption

pkg/deployment/resources/pdbs.go
```diff
@@ -0,0 +1,125 @@
...
+			pdb.Spec.MinAvailable = newFromInt(wantedMinAvail)
+
+			// Update the PDB
+			if err := pdbcli.Delete(pdbname, &metav1.DeleteOptions{}); err != nil {
+				log.Error().Err(err).Msg("PDB deletion failed")
+				maskAny(err)
+			}
+			// Create next reconcile loop, this should not be critical
+			// 	but only if wantedMinAvail is nonzero
+			return nil
```
ewoutp: This is unexpected. Do make sure to re-create here. Otherwise the name of the function `ensure...` is not fullfilled.

pkg/deployment/resources/pdbs.go
ewoutp: That is better than building your own helper that tweaks with structs

PR #355, opened 2019-03-12
Make cleanoutserver step mandatory in drain documentation.

neunhoef:
Internal Ref:
- https://github.com/arangodb/planning/issues/3733

### DIFF
--- a/docs/Manual/Deployment/Kubernetes/Drain.md
+++ b/docs/Manual/Deployment/Kubernetes/Drain.md
@@ -248,8 +248,8 @@ Here is a shell command which makes this check easy:
 curl -k https://arangodb.9hoeffer.de:8529/_db/_system/_api/replication/clusterInventory --user root: | jq . | grep '"isReady"\|"allInSync"' | sort | uniq -c
 ```
 
-If all these checks are performed and are okay, the cluster is ready to
-run a risk-free drain operation.
+If all these checks are performed and are okay, then it is safe to
+continue with the clean out and drain procedure as described below.
 
 {% hint 'danger' %}
 If there are some collections with `replicationFactor` set to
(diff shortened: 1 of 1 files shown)


docs/Manual/Deployment/Kubernetes/Drain.md
```diff
@@ -407,7 +421,7 @@ much data is stored in the pod, your mileage may vary, moving a terabyte
 of data can take considerably longer!
 
 If the optional step of
-[cleaning out a DBserver manually](#clean-out-a-dbserver-manually-optional)
+[cleaning out a DBserver manually](#clean-out-a-dbserver-manually)
 has been performed beforehand, the grace period can easily be reduced to 60
 seconds - at least from the perspective of ArangoDB, since the server is already
 cleaned out, so it can be dropped readily and there is still no risk.
```
Simran-B: @neunhoef It still says "optional" here. Not sure if the entire sentence should be rewritten...

neunhoef: It should, hang on...

neunhoef: Done.

Simran-B:
What is this sentence supposed to say?

> (please compare the above output of the `/_admin/cluster/health` API).

`compare above output to the output of ...` ?

Does the format follow the one of the cluster health API, or should the user check if both report the same things?

sleto-it:
@Simran-B to reply to your question:

I believe here we are asking the following:

- we are asking to call /_admin/cluster/cleanOutServer, and we have to pass to this call a json document with the name of the dbserver to be clened-out . In our example {"server":"DBServer0006"}

- "(please compare the above output of the /_admin/cluster/health API)" I believe means the following: we are saying to the user that this "DBServer0006" is not a name we invented - it is the one we get as "ShortName" from the /_admin/cluster/health

I believe this was the meaning - please could you reward and make it more clear?
Does this answer? so we merge and publish

thanks

PR #368, opened 2019-04-02
Feature/pod priority

maierlars:
Forwad pod priority class to pods.

### DIFF
--- a/docs/Manual/Deployment/Kubernetes/DeploymentResource.md
+++ b/docs/Manual/Deployment/Kubernetes/DeploymentResource.md
@@ -413,6 +413,10 @@ for each server of this group.
 This setting is not available for group `coordinators`, `syncmasters` & `syncworkers`
 because servers in these groups do not need persistent storage.
 
+### `spec.<group>.priorityClassName: string`
+
+Priority class name for pods of this group. Will be forwarded to the pod spec.
+
 ### `spec.<group>.probes.livenessProbeDisabled: bool`
 
 If set to true, the operator does not generate a liveness probe for new pods belonging to this group.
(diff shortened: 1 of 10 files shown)


manifests/templates/test/rbac.yaml
```diff
@@ -12,8 +12,8 @@ rules:
 - apiGroups: [""]
   resources: ["pods", "services", "persistentvolumes", "persistentvolumeclaims", "secrets", "serviceaccounts", "pods/log"]
   verbs: ["*"]
-- apiGroups: ["apps"]
-  resources: ["daemonsets", "deployments"]
+- apiGroups: ["apps", "scheduling.k8s.io"]
```
ewoutp: Separate apigroups in different rules. 1 api group per rule

docs/Manual/Deployment/Kubernetes/DeploymentResource.md
```diff
@@ -413,6 +413,10 @@ for each server of this group.
 This setting is not available for group `coordinators`, `syncmasters` & `syncworkers`
 because servers in these groups do not need persistent storage.
 
+### `spec.<group>.priorityClassName: string`
+
+Priority class name for pods of this group. Will be forwarded to the pod spec.
```
joerg84: Should we add a note that the class should be created upfront (potentially with link to k8s docu)?

pkg/util/k8sutil/pods.go
```diff
@@ -594,6 +599,9 @@ func CreateArangoSyncPod(kubecli kubernetes.Interface, developmentMode bool, dep
 	}
 	p.Spec.Containers = append(p.Spec.Containers, c)
 
+	// Add priorityClassName
+	p.Spec.PriorityClassName = podPriorityClassName
```
joerg84: Just curious: what is the behavior if you specify a class name of a class which doesn't exist? Do we need to handle something or emit a warning?

maierlars: Pod creation will fail and a event is attached to the arangodeployment resource containing a proper error message from kubernetes.

PR #369, opened 2019-04-03
Supported versions

sleto-it:
**do not merge yet**

---

**Fixes:**
- bare metal 1.10 is production ready

**Other:**
- adds entries for 1.11, 1.12, 1.13 and 1.14 for bare metals
- adds IBM Cloud
- adds operator version

Next steps (for **later**):
- the table should be part of the official doc - a link can be included in the repo readme and the actual content moved to doc/ for integration into the official manual
- we should reconsider 3.3.13 imho, and add a later version instead

Thanks,

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,23 +23,25 @@ it is intended to be.
 
 ### Production readiness state
 
-| Platform             | Kubernetes version | ArangoDB version | State | Production ready | Remarks               |
-|----------------------|--------------------|------------------|-------|------------------|-----------------------|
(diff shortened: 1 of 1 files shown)


joerg84:
Also, should we document a tested upgrade matrix (independent of this PR)?

joerg84:
@sleto-it Just curious what is the status here?

sleto-it:
@joerg84 joerg when we can have a call?

joerg84:
@sleto-it Pinged you on slack

PR #370, opened 2019-04-04
Health Call optimization

maierlars:
Reduce the amount of health calls to once every five seconds.


joerg84:
@maierlars  Is this still being worked on?

maierlars:
It is pending a review since Apr 4.

PR #373, opened 2019-04-04
waitForSync Upgrade

maierlars:
Monitor sync status concurrently. Only create new upgrade plan if everything is good.


pkg/deployment/reconcile/timeouts.go
```diff
@@ -33,5 +33,5 @@ const (
 	rotateMemberTimeout          = time.Minute * 30
 	shutdownMemberTimeout        = time.Minute * 30
 	upgradeMemberTimeout         = time.Hour * 6
-	waitForMemberUpTimeout       = time.Minute * 15
+	waitForMemberUpTimeout       = time.Minute * 45
```
neunhoef: Do we want this increase? What does the WaitForMemberUp action do now anyway? Let's discuss tomorrow.

PR #393, opened 2019-06-05
Added finalizer for log receiver.

maierlars:


maierlars:
Problem: The pre-stop hook waits for all finalizers to be removed. The log-receiver waits for the pod to be terminated before receiving the log and then removing the finalizer. Deadlock.

maierlars:
You fix that by ignoring the log-finalizer in the pre-stop hook. But for know, just leave it as is.

PR #408, opened 2019-06-25
ArangoSync-Client

maierlars:
Removed vendored directory. Use arangosync-client repo instead.


ewoutp:
External loading problem:

ewoutp:
I can now build against this branch

PR #413, opened 2019-07-04
Fixed required version of dependency.

joerg84:


joerg84:
It seems like the locked version is already at 4.1.2
```    "handlebars": {
      "version": "4.1.2",```

PR #425, opened 2019-07-29
Bug fix/unreliable cluster membership

ewoutp:
Fixed problem in MemberStatusList.Equal that causes the MemberOfCluster condition to be left out (in some cases).

It will also avoid updating the status of an ArangoDeployment in other cases.

### DIFF
--- a/pkg/apis/deployment/v1alpha/member_status_list_test.go
+++ b/pkg/apis/deployment/v1alpha/member_status_list_test.go
@@ -66,4 +66,17 @@ func TestMemberStatusList(t *testing.T) {
 	assert.Equal(t, m1.ID, (*list)[0].ID)
 	assert.Equal(t, m2.ID, (*list)[1].ID)
 	assert.Equal(t, m3.ID, (*list)[2].ID)
+
+	list2 := &MemberStatusList{m3, m2, m1}
+	assert.True(t, list.Equal(*list2))
+	assert.True(t, list2.Equal(*list))
+
+	list3 := &MemberStatusList{m3, m1}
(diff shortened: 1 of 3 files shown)


pkg/apis/deployment/v1alpha/member_status_list_test.go
```diff
@@ -66,4 +66,17 @@ func TestMemberStatusList(t *testing.T) {
...
+	assert.True(t, list2.Equal(*list))
+
+	list3 := &MemberStatusList{m3, m1}
+	assert.False(t, list.Equal(*list3))
+	assert.False(t, list3.Equal(*list))
+
+	list4 := MemberStatusList{m3, m2, m1}
+	list4[1].Phase = "something-else"
+	assert.False(t, list.Equal(list4))
+	assert.False(t, list4.Equal(*list))
```
joerg84: Do the following tests make sense? https://github.com/arangodb/kube-arangodb/pull/426

ewoutp: Merged them in here

pkg/deployment/resources/member_cleanup.go
```diff
@@ -72,6 +72,7 @@ func (r *Resources) cleanupRemovedClusterMembers() error {
 
 	// Only accept recent cluster health values
 	if time.Since(ts) > maxClusterHealthAge {
+		log.Info().Msg("Cleanup longer than max cluster health exiting")
```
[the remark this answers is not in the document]
ewoutp: done & added "age"

PR #438, opened 2019-08-12
Arango Backup support in Operator

ajanikow:


pkg/apis/deployment/v1alpha/backup_spec.go
```diff
@@ -0,0 +1,43 @@
...
+	Upload *ArangoBackupSpecOperation `json:"export,omitempty"`
+}
+
+type ArangoBackupSpecDeployment struct {
+	Name string `json:"name"`
+}
+
+type ArangoBackupSpecOperation struct {
+	RepositoryURL     string `json:"repositoryURL"`
+	CredentialsSecret string `json:"credentialsSecret,omitempty"`
```
maierlars: This should be `credentialsSecretName` to line up with other secret name fields.

pkg/apis/deployment/v1alpha/register.go
```diff
@@ -31,7 +31,11 @@ import (
 const (
 	ArangoDeploymentResourceKind   = "ArangoDeployment"
 	ArangoDeploymentResourcePlural = "arangodeployments"
-	groupName                      = "database.arangodb.com"
+
+	ArangoBackupResourceKind   = "ArangoBackup"
+	ArangoBackupResourcePlural = "arangobackups"
+
+	groupName = "database.arangodb.com"
```
maierlars: Maybe we should move the new code to `pkg/apis/backup/v1alpha1` to get correct semantic versioning.

pkg/backup/handlers/arango/backup/arango_client.go
```diff
@@ -0,0 +1,68 @@
...
+type ArangoClientFactory func(deployment *database.ArangoDeployment) (ArangoBackupClient, error)
+
+type ArangoBackupProgress struct {
+	Progress          int
+	Failed, Completed bool
+	FailMessage       string
+}
+
+type ArangoBackupClient interface {
+	Create() (driver.BackupMeta, error)
```
maierlars: `Create` needs more options ... [nitpicking]

pkg/apis/deployment/v1alpha/register.go
ajanikow: We cannot. Problem is that "deployment" directory is registering himself as "backup" in ClientSets. Changing this would create backward compatibility issues, which are not expected.

pkg/backup/handlers/arango/backup/state_create.go
```diff
@@ -0,0 +1,89 @@
...
+		backupMeta, err = client.Create()
+		if err != nil {
+			return switchTemporaryError(err, backup.Status)
+		}
+	}
+
+	details := &database.ArangoBackupDetails{
+		ID:                string(backupMeta.ID),
+		Version:           backupMeta.Version,
+		CreationTimestamp: now(),
```
maierlars: The backup itself will have a timestamp in its name. Those are likely to diverge but its okay for now. (List as known bug)

Makefile
```diff
@@ -165,7 +166,7 @@ update-generated:
 		"all" \
 		"github.com/arangodb/kube-arangodb/pkg/generated" \
 		"github.com/arangodb/kube-arangodb/pkg/apis" \
-		"deployment:v1alpha replication:v1alpha storage:v1alpha" \
+		"deployment:v1alpha replication:v1alpha storage:v1alpha \
```
ajanikow: Why removed quotation mark?

tests/backup_test.go
```diff
@@ -205,4 +231,38 @@ func TestBackupCluster(t *testing.T) {
...
+	t.Run("remove backup locally", func(t *testing.T) {
+		backup, name, id := ensureBackup(t, depl.GetName(), ns, deploymentClient, backupIsAvailable, nil)
+		defer deploymentClient.DatabaseV1alpha().ArangoBackups(ns).Delete(name, &metav1.DeleteOptions{})
+
+		// now remove the backup locally
+		err := databaseClient.Backup().Delete(nil, id)
+		assert.NoError(t, err, "Failed to delete backup: %s", err)
+
+		// wait for the backup to become unavailable
+		backup, err = waitUntilBackup(deploymentClient, backup.GetName(), ns, backupIsNotAvailable, 30*time.Second)
```
ajanikow: You should also check status - it should change to Deleted

pkg/backup/handlers/arango/backup/lifecycle.go
```diff
@@ -0,0 +1,36 @@
+package backup
```
joerg84: Copyright missing

pkg/backup/handlers/arango/backup/arango_client_impl.go
```diff
@@ -0,0 +1,241 @@
...
+		}, nil
+	}
+}
+
+func (ac *arangoClientBackupImpl) List() (map[driver.BackupID]driver.BackupMeta, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), defaultArangoClientTimeout)
+	defer cancel()
+
+	backups, err := ac.driver.Backup().List(ctx, nil)
+	if err != nil {
```
maierlars: This implementation is missing the temporary error detection. (just for the record)

pkg/backup/handlers/arango/backup/arango_client_impl.go
```diff
@@ -0,0 +1,241 @@
...
+	return raw, nil
+}
+
+func (ac *arangoClientBackupImpl) Upload(backupID driver.BackupID) (driver.BackupTransferJobID, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), defaultArangoClientTimeout)
+	defer cancel()
+
+	uploadSpec := ac.backup.Spec.Upload
+	if uploadSpec == nil {
+		return "", fmt.Errorf("upload was called but not upload spec was given")
```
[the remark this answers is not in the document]
Skarlso: Also, typo: `but *no Upload`...

pkg/backup/handlers/arango/backup/arango_client_impl.go
```diff
@@ -0,0 +1,241 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), defaultArangoClientTimeout)
+	defer cancel()
+
+	report, err := ac.driver.Backup().Progress(ctx, jobID)
+	if err != nil {
+		return ArangoBackupProgress{}, err
+	}
+
+	var ret ArangoBackupProgress
+	completedCount := 0
```
Skarlso: If the given default value is desired, use `var` for variable declaration. That will consistently set the value to its zero value.
```Go
var completedCount int
...
```

pkg/apis/backup/v1alpha/backup_status.go
```diff
@@ -0,0 +1,45 @@
...
+	Available         bool                 `json:"available"`
+}
+
+type ArangoBackupDetails struct {
+	ID                      string    `json:"id"`
+	Version                 string    `json:"version"`
+	PotentiallyInconsistent *bool     `json:"potentiallyInconsistent,omitempty"`
+	Uploaded                *bool     `json:"uploaded,omitempty"`
+	Downloaded              *bool     `json:"downloaded,omitempty"`
+	Imported                *bool     `json:"imported,omitempty"`
```
joerg84: Just curious: What is the reason for *bool here

pkg/backup/handlers/arango/backup/state_downloading.go
```diff
@@ -0,0 +1,98 @@
...
+		trueVar := true
+
+		return backupApi.ArangoBackupStatus{
+			Available:         true,
+			ArangoBackupState: newState(backupApi.ArangoBackupStateReady, "", nil),
+			Backup: &backupApi.ArangoBackupDetails{
+				ID:                string(backupMeta.ID),
+				Version:           backupMeta.Version,
+				CreationTimestamp: meta.Now(),
+				Downloaded:        &trueVar,
```
joerg84: Why not use the helper function https://github.com/arangodb/kube-arangodb/blob/cf338bd90649c65662ce6eedbc98fd84c50b8def/pkg/util/refs.go#L80

pkg/apis/backup/v1alpha/backup_status.go
ajanikow: It is one of the requirements - this mean Backup was created manually by hand and imported into ArangoBackup Resource Definition

pkg/backup/handlers/arango/backup/arango_client_impl.go
ajanikow: Temporary error detection is done - on the handler level

docs/Manual/Deployment/Kubernetes/BackupResource.md
```diff
@@ -0,0 +1,431 @@
+# ArangoBackup Custom Resource
+
+The ArangoDB Backup Operator creates and maintains ArangoDB Backups
+in a Kubernetes cluster, given a Backup specification.
+This deployment specification is a `CustomResource` following
+a `CustomResourceDefinition` created by the operator.
```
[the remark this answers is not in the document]
ajanikow: About restore pointed to ArangoDeployment docs.

pkg/apis/backup/v1alpha/backup_state.go
```diff
@@ -0,0 +1,78 @@
...
+	ArangoBackupStateCreate        state.State = "Create"
+	ArangoBackupStateUpload        state.State = "Upload"
+	ArangoBackupStateUploading     state.State = "Uploading"
+	ArangoBackupStateUploadError   state.State = "UploadError"
+	ArangoBackupStateReady         state.State = "Ready"
+	ArangoBackupStateDeleted       state.State = "Deleted"
+	ArangoBackupStateFailed        state.State = "Failed"
+)
+
+var ArangoBackupStateMap = state.Map{
```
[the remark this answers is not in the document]
ajanikow: It is just internal map to ensure states transitions

pkg/apis/backup/v1alpha/backup_status.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ArangoBackupStatus contains the status part of
+// an ArangoBackup.
+type ArangoBackupStatus struct {
+	ArangoBackupState `json:",inline"`
+	Backup            *ArangoBackupDetails `json:"backup,omitempty"`
+	Available         bool                 `json:"available"`
```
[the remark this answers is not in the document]
ajanikow: Available implies to state when we can restore from this backup (Ready, Upload statuses)

pkg/apis/backup/v1alpha/backup_validate.go
```diff
@@ -0,0 +1,81 @@
...
+	if err := a.Status.Validate(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (a *ArangoBackupSpec) Validate() error {
+	if a.Deployment.Name == "" {
+		return fmt.Errorf("deployment name can not be empty")
```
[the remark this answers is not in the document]
ajanikow: This change does not apply to linter, because of `Error strings should not be capitalized` error in IDE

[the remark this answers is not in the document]
Skarlso: That lint is there because they are trying to prevent double capitalization when the error message is printed out.

I'm not sure when and where Validate will be called, but looking at other things here, it looks like a lowercase version will be required because most of th time it's being appended to something or preceived by something, like a log message or formatting.

However, in that case @ajanikow you have to fix line 61. :) Which starts capitalized. :) Now I realize that in that case, RepositoryURL is a field, but so is Deployment.

Which means you have to decide now, if you want to say in the error more explicity something like this:

```go
fmt.Errorf("Deployment.Name can not be empty")
```

Which might be maybe preferred as it would indicate that a field name is being referenced here.

PR #447, opened 2019-09-20
Rename LocallyAttachedVolumes to NetworkAttachedVolumes.

neunhoef:
And change logic: default now false and meaning reversed.

### DIFF
--- a/docs/Manual/Deployment/Kubernetes/DeploymentResource.md
+++ b/docs/Manual/Deployment/Kubernetes/DeploymentResource.md
@@ -143,6 +143,19 @@ The encryption key cannot be changed after the cluster has been created.
 The secret specified by this setting, must have a data field named 'key' containing
 an encryption key that is exactly 32 bytes long.
 
+### `spec.networkAttachedVolumes: bool`
+
+The default of this option is `false`. If set to `true`, and the
+deployed ArangoDB version is new enough, a `ResignLeaderShip` operation
+will be triggered when a dbserver pod is evicted (rather than a
+`CleanOutServer` operation. Furthermore, the pod will simply be
+redeployed on a different node, rather than cleaned and retired and
+replaced by a new member. You must only set this option to `true` if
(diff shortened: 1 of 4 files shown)


docs/Manual/Deployment/Kubernetes/DeploymentResource.md
```diff
@@ -143,6 +143,19 @@ The encryption key cannot be changed after the cluster has been created.
 The secret specified by this setting, must have a data field named 'key' containing
 an encryption key that is exactly 32 bytes long.
 
+### `spec.networkAttachedVolumes: bool`
+
+The default of this option is `false`. If set to `true`, and the
+deployed ArangoDB version is new enough, a `ResignLeaderShip` operation
```
joerg84: Can we state the actual version? Also I would mention `You must only set this option to `true` if
your persistent volumes are "movable" in the sense that they can be
mounted from a different k8s node, like in the case of network attached
volumes. If your persistent volumes are tied to a specific pod, you
should leave this option on `false`.` First, as it at least for me helped me understand the relation to resignLeadership...

pkg/apis/deployment/v1alpha/deployment_spec.go
```diff
@@ -56,7 +56,7 @@ type DeploymentSpec struct {
 	DowntimeAllowed  *bool           `json:"downtimeAllowed,omitempty"`
 	DisableIPv6      *bool           `json:"disableIPv6,omitempty"`
 
-	LocallyAttachedVolumes *bool 	`json:"locallyAttachedVolumes,omitempty"`
+	NetworkAttachedVolumes *bool     `json:"networkAttachedVolumes,omitempty"`
```
joerg84: Is this a deprecation (i.e., did we ever release a version with LocallyAttachedVolumes)

neunhoef: No.

PR #463, opened 2019-10-04
Feature/change user password

informalict:


pkg/util/k8sutil/names.go
```diff
@@ -74,11 +74,9 @@ func stripArangodPrefix(id string) string {
 // complies with kubernetes name requirements.
 // If the name is to long or contains invalid characters,
 // if will be adjusted and a hash with be added.
-func FixupResourceName(name string, maxLength ...int) string {
```
maierlars: was `maxLength` actually unused? O_o

pkg/deployment/informers.go
```diff
@@ -143,7 +143,10 @@ func (d *Deployment) listenForSecretEvents(stopCh <-chan struct{}) {
 				}
 			},
 			UpdateFunc: func(oldObj, newObj interface{}) {
-				if _, ok := getSecret(newObj); ok {
```
maierlars: While this does work, it is quite unexpected here. 

To keep the design of the operator consistent, this logic should be placed in the inspection loop of the deployment, which is located at `pkg/deployment/deployment_inspector.go`.

pkg/util/k8sutil/names.go
informalict: Yes, it is unused.

pkg/deployment/informers.go
informalict: changed

pkg/deployment/resources/secret_hashes.go
```diff
@@ -78,6 +87,18 @@ func (r *Resources) ValidateSecretHashes() error {
 				Str("expected-hash", expectedHash).
 				Str("new-hash", hash).
 				Msg("Secret has changed.")
+			if actionHashChanged != nil {
+				if err := actionHashChanged(r.context, secret); err != nil {
+					log.Debug().Msg("Failed to change secret hash")
```
maierlars: better do a `failed to change secret: hash-changed-action returned error: ...`

pkg/deployment/resources/secret_hashes.go
```diff
@@ -78,6 +87,18 @@ func (r *Resources) ValidateSecretHashes() error {
...
+					return false, nil
```
maierlars: even if the action fails we may not want to stop handling the cluster and go into a `SecretChanged` state. I think the best would be if the action could decide if this change is a show stopper or not.

informalict: yes, the action function should decide if it is necessary to stop handling the cluster.
currently, it sets `SecretChanged` state when:
1) connection refuses to the database
2) cannot get a user 
   - if the user does not exist it creates the user
3) Cannot update a password for a user

Do you think I should omit one of the above conditions?

pkg/deployment/resources/secret_hashes.go
informalict: changed

pkg/deployment/resources/secret_hashes.go
informalict: > Code looks good to me. (Two little suggestions)
> 
> Please add a in-cluster test. Those tests are in the `/tests` folder. The test should verify that changing the password secret indeed changes the password in the database.

Integration test `TestSecretHashesRootUser` added. I could find function which checks if password in the database is changed. It checks if Status.SecretHashes.Users[root] is changed and it is changed when the password in the database is changed.

maierlars: You could try to connect with root user and the new password and ask for the version to see if the new password is valid.

maierlars: > Do you think I should omit one of the above conditions?

I think it never should stop handling the cluster. Not being able to change the password is not something that prevents me from chaos prevention.

pkg/apis/deployment/v1alpha/secret_hashes.go
```diff
@@ -34,11 +34,13 @@ type SecretHashes struct {
...
 	// SyncTLSCA contains the hash of the sync.tls.caSecretName secret
 	SyncTLSCA string `json:"sync-tls-ca,omitempty"`
+	// User's map contains hashes for each user
+	Users map[string]string `json:"users,omitempty"`
 }
 
 // Equal compares two SecretHashes
 func (sh *SecretHashes) Equal(other *SecretHashes) bool {
-	if sh == nil || other == nil {
+	if other == nil {
```
ajanikow: What if sh is nil? This check allows us to prevent NullPointerExceptions

informalict: changed

pkg/deployment/resources/secret_hashes.go
informalict: changed

informalict:
> During testing the following occurred:
> 
> 1. when the password secret is delete, the operator refuses to work. Instead nothing should happen.
> 2. after the bootstrap the root password is still empty and not as expected set the one in the secret. (current upstream master works fine)

Ad 1) Fixed
Ad 2) Checked and found one issue where there was no error handling for generating random root password

PR #466, opened 2019-10-11
Add test which checks if version has been changed

informalict:


tests/version_test.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		t.Fatalf("Deployment not running in time: %v", err)
+	}
+
+	single := depl.Status.Members.Single
+	if single == nil || len(single) == 0 {
+		t.Fatalf("single member is empty")
+	}
+
+	if string(single[0].ArangoVersion) != expectedVersion {
```
maierlars: I think we have special functions for comparing versions. Please have a look at the package containing the type of `ArangoVersion`.

PR #480, opened 2019-10-17
Docerize unit tests

informalict:
Improve unit tests

Remove unnecessary dependency

Discard constraints from dockerignore


informalict:
It is merged into the https://github.com/arangodb/kube-arangodb/pull/482

PR #486, opened 2019-10-21
Feature/add ubi based image

ajanikow:


maierlars:
So the UBI image is not built by default. I would prefer this to build by default. Because .. I know myself, I will forget that until the next release :D

Issue #437, opened 2019-08-08
When using VolumeClaimTemplates changing of StorageClass is not possible


maierlars:

informalict:
1) Two StorageClass had been created: storage (default), storage2.

2) Apply the below example into the kubernetes cluster:
 
apiVersion: "database.arangodb.com/v1alpha"
kind: "ArangoDeployment"
metadata:
  name: chaos
spec:
  imagePullPolicy: Always
  tls:
    caSecretName: None
    altNames: ["chaos.havefuninthe.net"]
  mode: Cluster
  locallyAttachedVolumes: false
  environment: Production
  license:
    secretName: arangodb-license-key
  image: arangodb/arangodb:3.3.17
  agents:
    count: 3
    storageClassName: standard
  dbservers:
    count: 2
    volumeClaimTemplate:
      spec:
        accessModes:
        - ReadWriteOnce
        resources:
          requests:
            storage: 8Gi
        volumeMode: Filesystem
    storageClassName: standard2
  coordinators:
    count: 3

3) Wait for five PVC (2 for dbservers, 3 for agents)

4) Change spec.dbservers.storageClassName to standard

5) Two dbservers have got new PVC created according to the new StorageClass (standard)

informalict:
Try with volumeClaimTemplate and storageClass inside this object

PR #487, opened 2019-10-21
Bug/change storage class name

informalict:
fixes https://github.com/arangodb/kube-arangodb/issues/437

### DIFF
--- a/pkg/deployment/cluster_scaling_integration.go
+++ b/pkg/deployment/cluster_scaling_integration.go
@@ -49,6 +49,10 @@ type clusterScalingIntegration struct {
 		arangod.NumberOfServers
 		mutex sync.Mutex
 	}
+	scaleEnabled struct {
+		mutex   sync.Mutex
+		enabled bool
+	}
 }
 
 const (
@@ -57,10 +61,12 @@ const (
 
 // newClusterScalingIntegration creates a new clusterScalingIntegration.
 func newClusterScalingIntegration(depl *Deployment) *clusterScalingIntegration {
-	return &clusterScalingIntegration{
+	ci := &clusterScalingIntegration{
 		log:  depl.deps.Log,
 		depl: depl,
 	}
+	ci.scaleEnabled.enabled = true
+	return ci
 }
 
 // SendUpdateToCluster records the given spec to be sended to the cluster.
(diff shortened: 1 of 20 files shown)


pkg/deployment/cluster_scaling_integration.go
```diff
@@ -253,3 +252,70 @@ func (ci *clusterScalingIntegration) updateClusterServerCount(ctx context.Contex
...
+func (ci *clusterScalingIntegration) EnableScalingCluster() error {
+	ci.scaleEnabled.mutex.Lock()
+	defer ci.scaleEnabled.mutex.Unlock()
+
+	if ci.scaleEnabled.enabled {
+		return nil
+	}
+
+	ctx := context.Background()
+	lastNumberOfServers := ci.GetLastNumberOfServers()
```
maierlars: I would prefer the operator to use the acutal values of running dbservers/coodrinators according to spec. This value here might be outdated

pkg/deployment/cluster_scaling_integration.go
```diff
@@ -253,3 +252,70 @@ func (ci *clusterScalingIntegration) updateClusterServerCount(ctx context.Contex
...
+	status, _ := ci.depl.GetStatus()
+	if !status.Plan.Empty() {
+		// Some Plan exists and scaling can not be enabled
+		ci.scaleEnabled.mutex.Unlock()
+		return false
+	}
+
+	// Scaling should be enabled because there is no Plan. It can happen when the action with enabling
+	ci.scaleEnabled.mutex.Unlock()
+	if ci.EnableScalingCluster() != nil {
```
maierlars: Here is a race! After line 316 in theory a new plan could be build and the `DisableScalingAction` could already been executed. Then this code would undo this change.

pkg/deployment/cluster_scaling_integration.go
informalict: changed

pkg/deployment/cluster_scaling_integration.go
informalict: changed

tests/persistent_volumes_test.go
```diff
@@ -258,3 +266,196 @@ func TestPVCTemplateResize(t *testing.T) {
...
+	})
+	require.NoError(t, err)
+
+	// Create deployment
+	_, err = arangoClient.DatabaseV1alpha().ArangoDeployments(k8sNameSpace).Create(depl)
+	require.NoError(t, err, "failed to create deployment: %s", err)
+	defer deferedCleanupDeployment(arangoClient, depl.GetName(), k8sNameSpace)
+
+	depl, err = waitUntilDeployment(arangoClient, depl.GetName(), k8sNameSpace, deploymentIsReady())
+	require.NoError(t, err, fmt.Sprintf("Deployment not running in time: %s", err)) // <-- fails here at the moment
```
maierlars: does this still fail?

tests/persistent_volumes_test.go
```diff
@@ -258,3 +266,196 @@ func TestPVCTemplateResize(t *testing.T) {
...
+	require.NotNil(t, depl.Spec.DBServers.VolumeClaimTemplate)
+
+	// Update deployment
+	depl.Spec.DBServers.VolumeClaimTemplate.Spec.StorageClassName = util.NewString(newStorageClassName)
+	_, err = arangoClient.DatabaseV1alpha().ArangoDeployments(k8sNameSpace).Update(depl)
```
maierlars: This looks unstable, because the operator could have updated the spec and thus reject the update because of an old revision? Just asking ...

tests/persistent_volumes_test.go
```diff
@@ -258,3 +266,196 @@ func TestPVCTemplateResize(t *testing.T) {
...
+			}
+		}
+		return nil
+	}
+
+	depl, err = waitUntilDeployment(arangoClient, depl.GetName(), k8sNameSpace, isStorageChanged, time.Minute*5)
+	require.NoError(t, err, "failed to change storage class for db servers: %s", err)
+
+	// Cleanup
+	removeDeployment(arangoClient, depl.GetName(), k8sNameSpace)
```
maierlars: I would love to see the following in addition:
Create a collection with 3 shards and put about 200 documents in there. Then start the test. After all volumes have been replaced, test if all the documents are still there. (They should ...)

tests/persistent_volumes_test.go
informalict: changed

tests/persistent_volumes_test.go
informalict: changed

tests/persistent_volumes_test.go
informalict: changed

PR #491, opened 2019-10-28
Add pod anti affinity

informalict:


chart/kube-arangodb/templates/deployment.yaml
```diff
@@ -46,6 +46,19 @@ spec:
                           operator: In
                           values:
                             - amd64
+              podAntiAffinity:
+                requiredDuringSchedulingIgnoredDuringExecution:
```
ajanikow: Please - change to preferredDuringSchedulingIgnoredDuringExecution

informalict: changed

PR #497, opened 2019-11-07
Drop all capabilities for security context

informalict:
IBM CV requires that created pods need to have dropped all capabilities.
Tested on arangodb/enterprise:3.5.2 with full integration test.

### DIFF
--- a/scripts/kube_create_license_key_secret.sh
+++ b/scripts/kube_create_license_key_secret.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 NS=$1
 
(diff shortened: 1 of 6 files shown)


scripts/kube_create_license_key_secret.sh
```diff
@@ -12,7 +12,14 @@ if [ -z "$2" ]; then
...
 
-LICENSE=$(echo -n "$2" | base64 -w 0)
+case $(uname) in
+    Darwin)
+        LICENSE=$(echo -n "$2" | base64 -b 0)
+        ;;
+    *)
+        LICENSE=$(echo -n "$2" | base64 -w 0)
+        ;;
+esac
```
maierlars: Thank you!

informalict: you are welcome :)

PR #499, opened 2019-11-11
Allow to set annotations in ArangoDeployment resources

ajanikow:
Add possibility to fix IBM CV `MeteringAnnotationsDefined`


chart/kube-arangodb/templates/deployment-operator/role.yaml
```diff
@@ -17,14 +17,14 @@ rules:
...
-      resources: ["pods", "services", "endpoints", "persistentvolumeclaims", "events", "secrets"]
+      resources: ["pods", "services", "endpoints", "persistentvolumeclaims", "events", "secrets", "serviceaccounts"]
       verbs: ["*"]
     - apiGroups: ["apps"]
       resources: ["deployments", "replicasets"]
       verbs: ["get"]
     - apiGroups: ["policy"]
       resources: ["poddisruptionbudgets"]
-      verbs: ["get", "create", "delete"]
+      verbs: ["*"]
```
maierlars: why do we need all?

ajanikow: We will need also "list", "update", "watch" - to be able to update/set annotations. Anyway, Operator manages PDB's for ArangoDeployments so it should be able to do all operations.

I can put other verbs here also, but it will be the same result as putting '*' there.

PR #501, opened 2019-11-22
Cleaning the code

informalict:
Unit tests:
I wrote 50 unit tests with different scenarios for `EnsurePods` and `ensureImages` and those tests pass on my branch and on the temporary branch `testing/after-cleaning` (I created from master that only for testing because I had to export a few functions and create short helpful functions, we can assume that branch is the same like master. Compare testing branch with master: https://github.com/arangodb/kube-arangodb/compare/testing/after-cleaning)

Integration tests:
Full integration tests passed with image `arangodb/enterprise-preview:3.5.3`

### DIFF
--- a/pkg/deployment/deployment_test.go
+++ b/pkg/deployment/deployment_test.go
@@ -1,3 +1,25 @@
+//
+// DISCLAIMER
+//
+// Copyright 2019 ArangoDB GmbH, Cologne, Germany
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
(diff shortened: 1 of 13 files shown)


pkg/deployment/deployment_test.go
```diff
@@ -0,0 +1,2327 @@
+package deployment
```
ajanikow: Add header

pkg/deployment/resources/pod_creator_arangod.go
```diff
@@ -0,0 +1,261 @@
+package resources
```
ajanikow: Add header

pkg/deployment/resources/pod_creator_sync.go
```diff
@@ -0,0 +1,208 @@
+package resources
```
ajanikow: Add header

pkg/util/k8sutil/lifecycle.go
```diff
@@ -0,0 +1,80 @@
+package k8sutil
```
ajanikow: Add header

pkg/deployment/deployment_test.go
informalict: added

pkg/deployment/resources/pod_creator_sync.go
informalict: added

pkg/util/k8sutil/lifecycle.go
informalict: added

PR #527, opened 2020-03-03
Remove v1alpha from core API

ajanikow:


joerg84:
I believe this entails some doc and example updates as well or?

PR #529, opened 2020-03-10
FEATURE/fix-linter-issues

ajanikow:


Makefile
```diff
@@ -157,13 +159,19 @@ allall: all
 # Tip: Run `eval $(minikube docker-env)` before calling make if you're developing on minikube.
 #
 
+GOLANGCI_ENABLED=deadcode gocyclo golint varcheck structcheck maligned errcheck \
+                 ineffassign interfacer unconvert goconst \
+                 megacheck
+
+#GOLANGCI_ENABLED+=dupl - disable dupl check
+
 .PHONY: fmt
```
informalict: I suggest calling that rule "linter" because it does more than "fmt"

ajanikow: Done

PR #533, opened 2020-03-20
[Feature] Add customizable volumes

ajanikow:


pkg/deployment/reconcile/plan_builder_storage.go
```diff
@@ -121,6 +121,11 @@ func createRotateServerStoragePlan(log zerolog.Logger, apiObject k8sutil.APIObje
 		}
 		return nil
 	})
+
+	if !plan.IsEmpty() {
```
informalict: that condition is unnecessary because we return `plan` variable anyway

PR #535, opened 2020-03-23
[Feature] Do not remove Agency members

ajanikow:
Difference between action `ActionTypeRemoveMember + ActionTypeAddMember` and `ActionTypeRecreateMember` is that action `ActionTypeRecreateMember` will not recreate PVC and Member ID - very useful in Ceerdinators to not create zombie Coordinators in cluster and for Agency to persist data in corner cases.

### DIFF
--- a/pkg/deployment/reconcile/action_recreate_member.go
+++ b/pkg/deployment/reconcile/action_recreate_member.go
@@ -63,7 +63,7 @@ func (a *actionRecreateMember) Start(ctx context.Context) (bool, error) {
 	_, err := a.actionCtx.GetPvc(m.PersistentVolumeClaimName)
 	if err != nil {
 		if kubeErrors.IsNotFound(err) {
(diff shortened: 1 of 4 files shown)


pkg/deployment/reconcile/action_recreate_member.go
```diff
@@ -63,7 +63,7 @@ func (a *actionRecreateMember) Start(ctx context.Context) (bool, error) {
 	_, err := a.actionCtx.GetPvc(m.PersistentVolumeClaimName)
 	if err != nil {
 		if kubeErrors.IsNotFound(err) {
-			return false, fmt.Errorf("PVC is missing %s. DBServer wont be recreated without old PV", m.PersistentVolumeClaimName)
+			return false, fmt.Errorf("PVC is missing %s. Member wont be recreated without old PV", m.PersistentVolumeClaimName)
```
joerg84: s/Member wont/Members won't

PR #550, opened 2020-04-14
Add istio friendly port names

sleepycat:
This commit changes the port names in the various service definitions to
follow the istio [naming convention](https://istio.io/docs/ops/configuration/traffic-management/protocol-selection/).
This helps the Kube-arangodb operator play nice with both istio and
observability tools like [Kiali](https://git.io/Jfesl).


ajanikow:
Hello!

This should be changed in helm chart, this manifests are generated from chart files. 

Port name can parametrized thru helm values.

Best Regards, 
Adam.

PR #551, opened 2020-04-14
Allow having one operator for all namespaces

couds:
This commit change the behavior of the operator to listen to all namespaces for the custom resources.

Before this commit you needed to install the deployment operator on each namespace you wanted to create a `ArangoDeployment`

### DIFF
--- a/manifests/arango-deployment.yaml
+++ b/manifests/arango-deployment.yaml
@@ -34,6 +34,24 @@ rules:
     - apiGroups: ["storage.k8s.io"]
       resources: ["storageclasses"]
       verbs: ["get", "list"]
+    - apiGroups: ["database.arangodb.com"]
+      resources: ["arangodeployments"]
+      verbs: ["*"]
+    - apiGroups: [""]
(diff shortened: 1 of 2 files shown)


ajanikow:
Hello!

We do not want to change default action taken by Operator - so it should be binded to namespace by default. We have security and lifecycle reasons to act like this - on kubernetes with multiple ArangoDB Deployments we can maintain multiple stages (prod, staging, dev) and upgrade them one by one. We are also sure that one broken Operator wont destroy other installation (RBAC).

If you need support like this then it should be implemented in generic way without changes in current default logic. You can add helm flag for mode, and based on this mode generate proper RBAC and deployment configs.

Best Regards, 
Adam.

rizzaricreative:
@ajanikow It seem like you are holding up the approval process with this change. This [pr](https://github.com/arangodb/kube-arangodb/pull/735) is waiting for you to review.

For most of us, a namespaced locked operator does not make sense neither does deploying it to multiple namespaces

PR #554, opened 2020-04-21
[Feature] Add optional Validation webhook

ajanikow:


chart/kube-arangodb-crd/templates/webhook/deployment.yml
```diff
@@ -0,0 +1,104 @@
...
+        release: {{ .Release.Name }}
+{{- if .Values.webhook.annotations }}
+      annotations:
+{{ toYaml .Values.webhook.annotations | indent 16 }}
+{{- end }}
+    spec:
+      serviceAccountName: {{ template "kube-arangodb-crd.name" . }}
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
```
joerg84, line 44: Why do we need this?

ajanikow: This will ensure that this container will run on amd64 only - it is recommended way how to do it (not by nodeSelector)

ajanikow:
It is valid for any version - our CRD and API is backward compatible, so we can validate Operator 0.4.0 API with Operator 1.1.0 API and it should work in same way.

pkg/util/log/factory.go
```diff
@@ -0,0 +1,49 @@
+//
+// DISCLAIMER
+//
+// Copyright 2020 ArangoDB GmbH, Cologne, Germany
```
informalict, line 4: 2020 -> 2021 in each file

informalict (commented):
LGTM

ajanikow:
Wont fix

PR #568, opened 2020-05-27
Add 'create' for backup operator role and remove namespace from cluster-wide resources

cristichiru:
Fixes backup operator complaining:

`E0527 13:24:20.300949       1 leaderelection.go:328] error initially creating leader election record: endpoints is forbidden: User "system:serviceaccount:arangodb:arango-backup-operator" cannot create resource "endpoints" in API group "" in the namespace "arangodb"`


ajanikow:
Hello!

This is fixed in change for kustomize support. Will be part of 1.1.0.

Best Regards,
Adam.

PR #619, opened 2020-08-18
Proposal: Alignment to Helm 3 commands and CRD management system

Glorf:
Thanks for your work on this operator, and on Arango as a whole.
I propose a POC set of (breaking) changes which enable operator builds using helm3. They also remove the need of separated helm chart for CRDs and use the new crds override system instead: https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you
Hope you'll make some use of these


Glorf:
Added new changes to increase flexibility of templating serviceAccount

orefalo:
neat initiative, was myself wondering why we had to use 3 helm commands to install arango in a cluster.

mysticaltech:
That would be great! And also support a proper Helm repo which is already implemented but not well supported.

ajanikow:
Covered already

PR #623, opened 2020-09-02
[Feature] Add Maintenance Mode support

ajanikow:


pkg/deployment/agency/definitions.go
```diff
@@ -26,4 +26,6 @@ const (
 	ArangoKey          = "arango"
 	PlanKey            = "Plan"
 	PlanCollectionsKey = "Collections"
+	PlanSupervision    = "Supervision"
+	PlanMaintenance    = "Maintenance2"
```
neunhoef, line 30: This looks odd, should it be `"Maintenance"` or is it needed at all?

neunhoef (approved):
LGTM. See one question.

pkg/deployment/agency/definitions.go
ajanikow: Obsolete, will cleanup it

PR #663, opened 2020-11-09
[Bugfix] Fix bootstrap phase

ajanikow:


pkg/deployment/reconcile/action_bootstrap_set_password.go
```diff
@@ -0,0 +1,148 @@
...
+		return "", maskAny(err)
+	}
+
+	password, err := a.ensureUserPasswordSecret(user, secret)
+	if err != nil {
+		return "", maskAny(err)
+	}
+
+	// Obtain the user
+	if u, err := client.User(context.TODO(), user); driver.IsNotFound(err) {
```
ewoutp, line 112: Why TODO instead of Background ?

ewoutp (approved):
I would prefer a lot more comments.
At least on all exposed types & functions.

I like that bootstrapping passwords is now a normal action.

pkg/deployment/reconcile/action_bootstrap_set_password.go
ajanikow: Will fix, was just used from other place - from k8s utils

PR #705, opened 2021-04-08
Tg64/requests timeout

informalict:


informalict (commented):
test

pkg/backup/handlers/arango/backup/arango_client_impl.go
```diff
@@ -128,9 +129,10 @@ func (ac *arangoClientBackupImpl) Get(backupID driver.BackupID) (driver.BackupMe
 	}
 }
 
-func (ac *arangoClientBackupImpl) getCredentialsFromSecret(secretName string) (interface{}, error) {
-
-	token, err := k8sutil.GetTokenSecret(ac.kubecli.CoreV1().Secrets(ac.backup.Namespace), secretName)
+func (ac *arangoClientBackupImpl) getCredentialsFromSecret(ctx context.Context, secretName string) (interface{}, error) {
+	ctxChild, cancel := context.WithTimeout(ctx, k8sutil.GetRequestTimeout())
+	token, err := k8sutil.GetTokenSecret(ctxChild, ac.kubecli.CoreV1().Secrets(ac.backup.Namespace), secretName)
+	cancel()
```
ajanikow, line 135: Why not just use defer cancel() line above?

pkg/deployment/context_impl.go
```diff
@@ -393,13 +405,17 @@ func (d *Deployment) DeletePod(podName string) error {
...
-func (d *Deployment) CleanupPod(p *v1.Pod) error {
+func (d *Deployment) CleanupPod(ctx context.Context, p *v1.Pod) error {
 	log := d.deps.Log
 	podName := p.GetName()
 	ns := p.GetNamespace()
 	options := meta.NewDeleteOptions(0)
 	options.Preconditions = meta.NewUIDPreconditions(string(p.GetUID()))
-	if err := d.deps.KubeCli.CoreV1().Pods(ns).Delete(context.Background(), podName, *options); err != nil && !k8sutil.IsNotFound(err) {
+	ctxChild, cancel := context.WithTimeout(ctx, k8sutil.GetRequestTimeout())
+	defer cancel()
```
ajanikow, line 415: Like here

pkg/deployment/deployment_inspector.go
```diff
@@ -45,10 +46,33 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	twentySeconds = time.Second * 20
```
ajanikow, line 50: Call it in more related to use-case way, like PerNodeReconcilationTimeout

pkg/deployment/deployment_inspector.go
```diff
@@ -59,13 +83,22 @@ var (
...
+
+	timeout, err := d.getReconciliationTimeout()
+	if err != nil {
+		log.Error().Err(err).Msg("Unable to get nodes")
+		return minInspectionInterval // Retry ASAP
+	}
+
+	ctxReconciliation, cancelReconciliation := context.WithTimeout(context.Background(), timeout)
 	defer func() {
+		cancelReconciliation()
```
ajanikow, line 95: Use 2 defers, make it consistent across PR

ajanikow (changes requested):
Please, make ctx cancel management consistent across PR

pkg/backup/handlers/arango/backup/arango_client_impl.go
informalict: yes, it is the place where `defer` can be used. My approach is:
- if the timeout is used only once then `defer` can be used
- otherwise cancel context immediately and reuse the same cancel function later in the same function

pkg/deployment/deployment_inspector.go
informalict: changed

pkg/deployment/deployment_inspector.go
informalict: changed

ajanikow (approved):

PR #707, opened 2021-04-13
[Bugfix] Fix Maintenance Action plan

ajanikow:


pkg/deployment/reconcile/plan_builder_common.go
```diff
@@ -65,7 +65,7 @@ func createMaintenanceManagementPlan(ctx context.Context,
 
 	if m.Enabled() && !spec.Database.GetMaintenance() {
 		log.Info().Msgf("Disabling maintenance mode")
```
informalict, line 67: Disabling maintenance mode -> Enabling maintenance mode

informalict (approved):

PR #726, opened 2021-05-17
Patch arango-local-storage.yaml

edgarlanting:
Fix a typo/error from `apiVersion: "storage.arangodb.com/v1"` to `apiVersion: "storage.arangodb.com/v1alpha"`


ajanikow:
Storate API is still in v1alpha version

PR #728, opened 2021-05-18
[Feature] Cluster IP communication

ajanikow:


pkg/deployment/pod/utils.go
```diff
@@ -0,0 +1,68 @@
...
+			switch method {
+			case api.DeploymentCommunicationMethodDNS:
+				return k8sutil.CreateServiceDNSNameWithDomain(svc, spec.ClusterDomain), nil
+			case api.DeploymentCommunicationMethodIP:
+				if svc.Spec.ClusterIP == "" {
+					return "", errors.Newf("ClusterIP of service %s is empty", svc.GetName())
+				}
+
+				return svc.Spec.ClusterIP, nil
+			case api.DeploymentCommunicationMethodShortDNS:
```
informalict, line 57: it is deadcode because the `method` can be only api.DeploymentCommunicationMethodDNS or api.DeploymentCommunicationMethodIP and the below `default` is also the deadcode. I think that we should use only one `switch` and don't put `switch` into  `switch`

pkg/deployment/reconcile/action_wait_for_member_up.go
```diff
@@ -150,6 +151,8 @@ func (a *actionWaitForMemberUp) checkProgressAgent(ctx context.Context) (bool, b
 	shortCtx, c := context.WithTimeout(ctx, 3*time.Second)
 	defer c()
 
+	shortCtx = agency.WithAllowDifferentLeaderEndpoints(ctx)
```
informalict, line 154: ctx -> shortCtx

informalict (changes requested):

informalict (approved):

Issue #743, opened 2021-06-30
Operator is trying to get pods on cluster scope


duduz:
Hi Everyone,

We have observed the following behavior on one of clusters where the operator is failing to update ArangoDeployment during deployment inspection
When it fails, it constantly trying to list the pods on the cluster scope. and become sort of "stuck in a loop" trying to get the pods from the cluster scope. which results in the operator become non-responsive.
upon restart of the operator pod, this no longer occurs.

### Environment
kubernetes 1.19.7 via AKS.
kube-arangodb operator version is 1.15

### Logs

Date | Event
------------ | -------------
6/15/2021,   10:08:43.121 AM | 2021-06-15T10:08:43Z   DBG Inspect deployment... component=deployment deployment=arangodb-cluster   operator-id=ssk6j
6/15/2021, 10:08:43.214 AM | 2021-06-15T10:08:43Z WRN Member is   not ready for long time, but it is not safe to mark it a failed because: TODO   component=deployment deployment=arangodb-cluster id=SNGL-ben2cwi1   operator-id=ssk6j role=single
6/15/2021, 10:08:43.220 AM | 2021-06-15T10:08:43Z INF Ensuring   labels component=deployment deployment=arangodb-cluster operator-id=ssk6j
6/15/2021, 10:08:43.220 AM | 2021-06-15T10:08:43Z INF Ensuring   annotations
6/15/2021, 10:08:43.266 AM | 2021-06-15T10:08:43Z INF   Reconciliation loop took 76.365279ms component=deployment   deployment=arangodb-cluster operator-id=ssk6j
6/15/2021, 10:08:43.266 AM | 2021-06-15T10:08:43Z DBG   ...inspected deployment component=deployment deployment=arangodb-cluster   interval=10s operator-id=ssk6j
6/15/2021, 10:08:43.266 AM | 2021-06-15T10:08:43Z INF Inspect   loop took 144.685529ms component=deployment deployment=arangodb-cluster   operator-id=ssk6j
6/15/2021, 10:08:53.000 AM | 2021-06-15T10:08:53Z DBG Inspect   deployment... component=deployment deployment=arangodb-cluster   operator-id=ssk6j
[...]

PR #744, opened 2021-07-01
[Bugfix] Use persistent name and namespace

ajanikow:
Persist name and namespace of deployment to prevent issues when K8S API cannot be contacted


ajanikow:
Fixes: https://github.com/arangodb/kube-arangodb/issues/743

ewoutp (approved):
LGTM

PR #752, opened 2021-07-08
[Feature] Version Endpoint

ajanikow:


Makefile
```diff
@@ -30,6 +30,8 @@ GOVERSION := 1.10.0-alpine
 PULSAR := $(GOBUILDDIR)/bin/pulsar$(shell go env GOEXE)
 GOASSETSBUILDER := $(GOBUILDDIR)/bin/go-assets-builder$(shell go env GOEXE)
 
+BUILDTIME = $(shell go run "$(ROOT)/tools/dategen/")
```
ajanikow, line 33: Make evaluate this variable when it is called, so binary will be executed only durig build

ewoutp (approved):
LGTM

PR #756, opened 2021-07-14
Replace deprecated node labels

### LINKED ISSUE #751: Operator not deployable on GKE autopilot
Hello, I tried to deploy the operator and a database onto a GKE autopilot cluster. However, the manifests for the operator and for the databases it generates are blocked by the gatekeeper validation that comes on an autopilot cluster.

Specifically, the `nodeAffinity` rules that use `beta.kubernetes.io/arch` is not allowed:

```
Normal   New Single Added      2m33s                 arango-deployment-operator-64bdb48575-nfc6r  New single SNGL-vsxodwym added to deployment
Warning  Reconcilation Failed  92s (x24 over 2m20s)  arango-deployment-operator-64bdb48575-nfc6r  Image detection failed: admission webhook "validation.gatekeeper.sh" denied the request: [denied by autogke-node-affinity-selector-limitation] If not using workload separation, node affinity is not allowed on labels with keys: <{"beta.kubernetes.io/arch"}>; Autopilot allows node affinity only on labels with keys: <["topology.kubernetes.io/region", "topology.kubernetes.io/zone", "failure-domain.beta.kubernetes.io/region", "failure-domain.beta.kubernetes.io/zone", "cloud.google.com/gke-os-distribution", "kubernetes.io/os", "kubernetes.io/arch"]>. Minimum required cpu request for using workload separation: <500>m, current pod request : <500>m. Requesting user: <system:serviceaccount:default:arango-deployment-operator> and groups: <["system:serviceaccounts", "system:serviceaccounts:default", "system:authenticated"]>
```

I also had this error when installing the operator, but was able to remove the node affinity rule in this instance.

I'm not sure why this architecture based affinity rule is there in the first place, other than to ensure the operator is only running on amd64 nodes. However, this could be done via documentation stating that the container images are only for amd64, rather than adding this by default to the deployment manifests.

Due to the way Autopilot is configured, I cannot remove this rule, so currently nobody using GKE autopilot will be able to install the operator.

davidsbond:
Closes #751

This commit updates all references to `beta.kubernetes.io/arch` with `kubernetes.io/arch`. The `beta` prefixed
label is [deprecated](https://kubernetes.io/docs/reference/labels-annotations-taints/#beta-kubernetes-io-arch-deprecated) and causes issues when trying to deploy onto a GKE autopilot cluster which maintains strict rules around which node labels can be used for affinity.

Signed-off-by: David Bond <davidsbond93@gmail.com>

### DIFF
--- a/chart/arangodb-ingress-proxy/templates/deployment.yaml
+++ b/chart/arangodb-ingress-proxy/templates/deployment.yaml
@@ -31,7 +31,7 @@ spec:
                     requiredDuringSchedulingIgnoredDuringExecution:
                         nodeSelectorTerms:
                             - matchExpressions:
-                                  - key: beta.kubernetes.io/arch
+                                  - key: kubernetes.io/arch
                                     operator: In
                                     values:
                                         - amd64
--- a/chart/kube-arangodb/templates/deployment.yaml
+++ b/chart/kube-arangodb/templates/deployment.yaml
@@ -56,7 +56,7 @@ spec:
                 requiredDuringSchedulingIgnoredDuringExecution:
                   nodeSelectorTerms:
                     - matchExpressions:
-                        - key: beta.kubernetes.io/arch
+                        - key: kubernetes.io/arch
                           operator: In
                           values:
                             - amd64
--- a/manifests/arango-all.yaml
+++ b/manifests/arango-all.yaml
@@ -478,7 +478,7 @@ spec:
     replicas: 2
     strategy:
         type: Recreate
-        
+
     selector:
         matchLabels:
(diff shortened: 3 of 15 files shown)


owenmonagan:
Is this going to be merged? I would also love to use an GKE autopilot cluster

davidsbond:
@owenmonagan, based on https://github.com/arangodb/kube-arangodb/issues/751#issuecomment-880581389 I don't think so, looks like they'll likely be implementing a half-way version where you can specify the labels yourself. Doesn't look like it made it into 1.2.1

PR #759, opened 2021-07-20
Feature/disable support for eol arangodb 3.5 versions

informalict:


pkg/deployment/resources/pod_termination.go
```diff
@@ -191,12 +181,6 @@ func (r *Resources) prepareDBServerPodTermination(ctx context.Context, log zerol
 		dbserverDataWillBeGone = true
 	}
 
-	// Is this a simple pod restart?
```
informalict, line 194: @ajanikow here please pay special attention.

PR #770, opened 2021-08-03
Graceful shutdown for coordinators

informalict:


ajanikow (changes requested):
Changelog entry missing

ajanikow (approved):

pkg/deployment/reconcile/helper_shutdown.go
```diff
@@ -105,12 +114,43 @@ func (s shutdownHelperAPI) CheckProgress(ctx context.Context) (bool, bool, error
...
+	}
+
+	info, err := c.ShutdownInfoV2(ctx)
+	if err != nil {
+		log.Debug().Err(err).Msg("Failed to fetch shutdown info from the coordinator")
+		return false, false, errors.WithStack(err)
+	}
+
+	// Log the status of the shutdown process.
+	log.Info().
```
ajanikow, line 140: Do you want to log it each 100ms?

pkg/deployment/reconcile/helper_shutdown.go
```diff
@@ -75,10 +77,17 @@ func (s shutdownHelperAPI) Start(ctx context.Context) (bool, error) {
 			return false, errors.WithStack(err)
 		}
 		removeFromCluster := false
-		log.Debug().Bool("removeFromCluster", removeFromCluster).Msg("Shutting down member")
+		gracefulShutdown := false
+		if isGracefulShutdownAvailable(group, m.ArangoVersion) {
```
ajanikow, line 81: Is this needed? Anyway this param should be ignored on older versions.

pkg/deployment/reconcile/helper_shutdown.go
```diff
@@ -105,12 +114,43 @@ func (s shutdownHelperAPI) CheckProgress(ctx context.Context) (bool, bool, error
 		log.Error().Msg("No such member")
 		return true, false, nil
 	}
-	if !m.Conditions.IsTrue(api.ConditionTypeTerminated) {
-		// Pod is not yet terminated
+	if m.Conditions.IsTrue(api.ConditionTypeTerminated) {
```
ajanikow, line 117: Make it simple, just wait for shutdown.

ajanikow (changes requested):
What about kubectl removal of pods? You can solve it with finalizers, to take shutdown as finalizer in coordinator. PreStop hook will anyway stop shutdown until finalizer is removed.

Also, verify what will happen if you will send 2 requests of shutdown (so we need to do check in finalizer if shutdown is in process)

pkg/deployment/reconcile/helper_shutdown.go
informalict: It is required because a graceful shutdown is available from the specific version and only it applies to the coordinator. Everything is checked in the function `isGracefulShutdownAvailable`

pkg/deployment/reconcile/helper_shutdown.go
informalict: removed

pkg/deployment/reconcile/helper_shutdown.go
informalict: changed

pkg/deployment/reconcile/helper_shutdown.go
informalict: removed

pkg/deployment/reconcile/helper_shutdown.go
```diff
@@ -74,11 +77,24 @@ func (s shutdownHelperAPI) Start(ctx context.Context) (bool, error) {
...
+		gracefulShutdown := true
+
+		log.Debug().
+			Bool("removeFromCluster", removeFromCluster).
+			Bool("gracefulShutdown", gracefulShutdown).
+			Msg("Shutting down member")
 		ctxChild, cancel = context.WithTimeout(ctx, shutdownTimeout)
 		defer cancel()
-		if err := c.Shutdown(ctxChild, removeFromCluster); err != nil {
+		if err := c.ShutdownV2(ctxChild, removeFromCluster, gracefulShutdown); err != nil {
```
ajanikow, line 97: Hardcode to true

ajanikow (changes requested):
Just few small comments

ajanikow:
Also, add finalizer handler (send shutdown request) here https://github.com/arangodb/kube-arangodb/blob/master/pkg/deployment/resources/pod_finalizers.go#L81

pkg/deployment/reconcile/helper_shutdown.go
```diff
@@ -74,11 +77,24 @@ func (s shutdownHelperAPI) Start(ctx context.Context) (bool, error) {
...
 		}
+
+		if isShutdownOngoing, err := isShutdownOngoing(ctx, c); err != nil {
+			log.Debug().Err(err).Msg("Failed to check shutdown info")
+			return false, errors.WithStack(err)
+		} else if isShutdownOngoing {
+			return true, nil
+		}
+
 		removeFromCluster := false
```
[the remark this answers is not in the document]
informalict: removed

pkg/deployment/reconcile/helper_shutdown.go
informalict: changed

pkg/deployment/reconcile/helper_shutdown.go
```diff
@@ -74,11 +77,24 @@ func (s shutdownHelperAPI) Start(ctx context.Context) (bool, error) {
...
-		log.Debug().Bool("removeFromCluster", removeFromCluster).Msg("Shutting down member")
+		gracefulShutdown := true
```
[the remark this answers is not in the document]
informalict: removed

PR #774, opened 2021-08-18
[Feature] High priority plan

ajanikow:


pkg/deployment/reconcile/plan_executor.go
```diff
@@ -38,28 +38,108 @@ import (
 	"github.com/arangodb/kube-arangodb/pkg/util/k8sutil"
 )
 
+type planGetter interface {
```
informalict, line 41: planGetter gets and sets so maybe to call it `planner`?
if so then change planGetterNormal -> planNormal and planGetterHigh -> planHigh.
I think we can describe it that we have different `planners` and two implementations: planNormal, planHigh

informalict (changes requested):

informalict (approved):
LGTM

PR #794, opened 2021-09-22
[Feature] SchedulerName Pod Customization

ajanikow:


informalict (approved):
LGTM

pkg/deployment/rotation/arangod.go
```diff
@@ -0,0 +1,37 @@
...
+package rotation
+
+import (
+	api "github.com/arangodb/kube-arangodb/pkg/apis/deployment/v1"
+	core "k8s.io/api/core/v1"
+)
+
+func podCompare(_ api.DeploymentSpec, _ api.ServerGroup, spec, status *core.PodSpec) compareFunc {
+	return func(builder api.ActionBuilder) (mode Mode, plan api.Plan, err error) {
+		if spec.SchedulerName != status.SchedulerName {
```
informalict, line 30: you should compare values , not pointers

ajanikow: It is "string", not "*string"

PR #795, opened 2021-09-23
[Feature] Enterprise Edition support

ajanikow:


Makefile
```diff
@@ -95,11 +97,13 @@ endif
 
 BASEUBIIMAGE ?= registry.access.redhat.com/ubi8/ubi-minimal:8.4
 
+OPERATORIMAGENAME ?= $(REPONAME)
```
ewoutp, line 100: Is this overwritten for enterprise?

Robert-Stam (approved):
LGTM

Makefile
ajanikow: Yes, release-manager will set it for enterprise to (kube-arangodb-enterprise or kube-arangodb-enterprise-preview)

ajanikow: In make we just set default value to be able to build everything using old methods

ewoutp (approved):
LGTM

PR #804, opened 2021-10-04
[Feature] Topology Support

ajanikow:


informalict (approved):
Disclaimer does not contain the author almost in each file.
The year in the disclaimer should be 2021 not 2020.

PR #828, opened 2021-11-08
[Feature] Refactor Pod and containter creator part2

informalict:
add 4 new functions to the interfaces:
GetArgs
GetName
GetVolume
GetVolumeMounts

### DIFF
--- a/pkg/deployment/resources/pod_creator_arangod.go
+++ b/pkg/deployment/resources/pod_creator_arangod.go
@@ -66,8 +66,8 @@ type MemberArangoDPod struct {
 	resources        *Resources
 	imageInfo        api.ImageInfo
 	autoUpgrade      bool
-	id               string
 	args             []string
+	volumes          pod.Volumes
 }
 
 type ArangoDContainer struct {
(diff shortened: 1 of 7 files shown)


pkg/deployment/resources/pod_creator_arangod.go
```diff
@@ -66,7 +66,8 @@ type MemberArangoDPod struct {
 	resources        *Resources
 	imageInfo        api.ImageInfo
 	autoUpgrade      bool
-	id               string
+	args             []string
```
ajanikow, line 69: Why providing it as static list? We can generate it

pkg/deployment/resources/pod_creator_arangod.go
```diff
@@ -66,7 +66,8 @@ type MemberArangoDPod struct {
...
+	volumes          pod.Volumes
```
ajanikow, line 70: Also here, I would vote to generate it on demand

ajanikow (changes requested):

pkg/deployment/resources/pod_creator_arangod.go
informalict: changed

pkg/deployment/resources/pod_creator_arangod.go
informalict: changed

ajanikow (approved):

PR #830, opened 2021-11-09
Enable metrics exporter to work without authorization

Strangerxxx:
Before it metrics require arangodb jwt token authorization
Now authorization is not necessary for metrics exporter


ajanikow:
Fixed in 1.2.6

PR #834, opened 2021-11-11
Arangojob

jwierzbo:


Dockerfile
```diff
@@ -1,6 +1,8 @@
-ARG IMAGE=scratch
```
jwierzbo, line 1: this will be reverted

chart/kube-arangodb/templates/apps-operator/role.yaml
```diff
@@ -0,0 +1,54 @@
+{{ if .Values.rbac.enabled -}}
+{{ if .Values.operator.features.backup -}}
```
jwierzbo, line 2: this will be changed into `features.apps`

informalict (approved):
LGTM

PR #847, opened 2021-12-03
go1.7 adjustment

jwierzbo:
Due to following error on Mac arm64 arch (M1 CPU) `combining dwarf failed: unknown load command`, Go 1.17 is required since below fix is not available yet for Go 1.16:
https://github.com/golang/go/issues/47269


jwierzbo:
we are not going to support Go 1.17 yet

PR #848, opened 2021-12-03
Debug mode for DLV debugger

jwierzbo:


.dockerignore
```diff
@@ -1,6 +1,5 @@
 .DS_Store
 .git
-.gobuild
```
ajanikow, line 3: Why do you remove this?

Dockerfile.debug
```diff
@@ -0,0 +1,43 @@
+ARG RELEASE_MODE=community
+ARG TARGETARCH=amd64
+
+FROM golang:1.16 as builder
```
informalict, line 4: maybe we should consider to add the below line before FROM:
ARG GO_VERSION=1.16 
and use it in FROM

informalict (approved):
LGTM

PR #850, opened 2021-12-06
[Feature] Agency Cache

ajanikow:


jwierzbo (approved):

pkg/deployment/agency/state.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	state := c[0].Arango
+
+	if _, ok := state.Current.Collections["_system"]; !ok {
+		return State{}, errors.Newf("Unable to find system database (invalid data)")
+	}
+
+	if _, ok := state.Plan.Collections["_system"]; !ok {
+		return State{}, errors.Newf("Unable to find system database (invalid data)")
```
informalict, line 74: Unable to find system database (invalid data) -> Unable to find system database in the plan (invalid data)

pkg/deployment/agency/state.go
```diff
@@ -0,0 +1,117 @@
+//
+// DISCLAIMER
+//
+// Copyright 2016-2021 ArangoDB GmbH, Cologne, Germany
```
informalict, line 4: why 2016. it is a new file and should start with 2021

informalict (approved):
LGTM with two comments

pkg/deployment/agency/state.go
ajanikow: We use everywhere licenses like this

pkg/deployment/agency/state.go
ajanikow: This error should never happen, unless agency is not totally broken

PR #857, opened 2021-12-10
[Feature] Allow to disable ClusterScalingIntegration and add proper S…

ajanikow:
…cheduled label to pods

### DIFF
--- a/pkg/apis/deployment/v1/topology_status.go
+++ b/pkg/apis/deployment/v1/topology_status.go
@@ -103,6 +103,28 @@ func (t *TopologyStatus) IsTopologyOwned(m *TopologyMemberStatus) bool {
 	return t.ID == m.ID
 }
 
+func (t *TopologyStatus) IsTopologyEvenlyDistributed(group ServerGroup) bool {
+	if t == nil {
+		return true
+	}
+
+	max, min := 0, math.MaxInt64
+
+	for _, z := range t.Zones {
+		l := len(z.Members[group.AsRoleAbbreviated()])
+
+		if min > l {
+			min = l
+		}
+
+		if max < l {
+			max = l
+		}
+	}
+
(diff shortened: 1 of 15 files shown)


jwierzbo (approved):

pkg/apis/deployment/v1/topology_status.go
```diff
@@ -103,6 +103,28 @@ func (t *TopologyStatus) IsTopologyOwned(m *TopologyMemberStatus) bool {
 	return t.ID == m.ID
 }
 
+func (t *TopologyStatus) IsTopologyEvenlyDistributed(group ServerGroup) bool {
```
informalict, line 106: I would like to have a description here. What function should return when `t.Zones` is empty. Currently, 
the function `IsTopologyEvenlyDistributed` returns false (not distributed evenly) when t.Zones are empty. Is it intentional? The description of the function should explain it because I don't know the intensions

pkg/deployment/resources/pod_inspector.go
```diff
@@ -184,6 +186,20 @@ func (r *Resources) InspectPods(ctx context.Context, cachedStatus inspectorInter
 			}
 		}
 
+		if k8sutil.IsPodScheduled(pod) {
+			if _, ok := pod.Labels[k8sutil.LabelKeyArangoScheduled]; k8sutil.IsPodScheduled(pod) && !ok {
```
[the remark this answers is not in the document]
ajanikow: Will fix it - small issue which just use cycles

pkg/apis/deployment/v1/topology_status.go
ajanikow: And t.Zones cant be with size 0 - it is checked in code. If it is, then false is expected

PR #901, opened 2022-02-03
[Cleanup] Reorganize main reconciliation context

ajanikow:


pkg/deployment/server_api.go
```diff
@@ -44,8 +44,8 @@ func (d *Deployment) Namespace() string {
 	return d.apiObject.Namespace
 }
 
-// Mode returns the mode of the deployment.
-func (d *Deployment) Mode() api.DeploymentMode {
+// GetAgencyClientsWithPredicate returns the mode of the deployment.
```
informalict, line 47: GetMode instead GetAgencyClientsWithPredicate

informalict (approved):
LGTM with small comment

pkg/deployment/server_api.go
nikita-vanyasin, line 47: I guess there should be GetMode instead

nikita-vanyasin (changes requested):
All LGTM except small comment fix for `GetMode()` function.

These new TODO comments in tests at [235fc6f789315bea8d9d18981b399a7c880bac8cc38421cfba0a91a818e96656R97](https://github.com/arangodb/kube-arangodb/pull/901/files#diff-235fc6f789315bea8d9d18981b399a7c880bac8cc38421cfba0a91a818e96656R98)
  also does look a bit confusing but I guess this is some old code moved from one place to another?

nikita-vanyasin (approved):

ajanikow:
> All LGTM except small comment fix for `GetMode()` function.
> 
> These new TODO comments in tests at [235fc6f789315bea8d9d18981b399a7c880bac8cc38421cfba0a91a818e96656R97](https://github.com/arangodb/kube-arangodb/pull/901/files#diff-235fc6f789315bea8d9d18981b399a7c880bac8cc38421cfba0a91a818e96656R98) also does look a bit confusing but I guess this is some old code moved from one place to another?

We are not implementing not-used test functions - it is desired to do panic with them to make sure that they are not used.

Issue #887, opened 2022-01-12
Unreliable finalizers


Sieabah:
I have found this operator to be okay at keeping things running from a complete node failure. If for whatever reason it decides it needs to evict any pod from one node to another however it gets stuck in the drain finalizer indefinitely. It will never resolve under any circumstance. I have to manually remove the drain finalizer in order for the operator to recover and correctly setup the pod again.

ajanikow:
Hello!

The latest Operator version does not use a drain finalizer anymore, now a graceful one is used (which tolerates a little bit more node failures).

Also, please set: https://www.arangodb.com/docs/stable/deployment-kubernetes-deployment-resource.html#specnetworkattachedvolumes-bool to true, it will resolve your issue in the current version (we will change the default in 1.3.0).

Best Regards,
Adam.

Sieabah:
@ajanikow The issue one Azure is that I've had issues with the attached volumes being unmounted reliably.

Will there be a build soon of 1.3.0? I only see 1.2.6 being available.

ajanikow:
Hello!

You can change it even in 1.2.6 to true in ArangoDeployment, you do not have to wait for 1.3.0.

Best Regards,
Adam.

Sieabah:
The issue is still with the finalizers, 

    finalizers:
        - database.arangodb.com/graceful-shutdown

Has been stuck on terminating for 8 hours.

ajanikow:
Hello!

Can you share logs and try to use 1.2.7? In 1.2.7 it should be less graceful.

Also, can you share yaml of this pod?

Best Regards,
Adam.

Sieabah:
Here are the two pod YAMLs that are perpetually terminating

```
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: "2022-01-19T20:01:28Z"
  deletionGracePeriodSeconds: 0
  deletionTimestamp: "2022-01-19T20:29:52Z"
  finalizers:
  - database.arangodb.com/graceful-shutdown
  labels:
    app: arangodb
    arango_deployment: prod-arango
    deployment.arangodb.com/member: PRMR-rvytgxck
    deployment.arangodb.com/scheduled: "1"
    role: dbserver
  name: prod-arango-prmr-rvytgxck-3f9aff
  namespace: production
  ownerReferences:
  - apiVersion: database.arangodb.com/v1
    controller: true
    kind: ArangoDeployment
    name: prod-arango
[...]

PR #902, opened 2022-02-03
[Bugfix] Add unreachable condition

ajanikow:


pkg/deployment/resources/pod_inspector.go
```diff
@@ -218,6 +218,20 @@ func (r *Resources) InspectPods(ctx context.Context, cachedStatus inspectorInter
...
 		// End of Topology labels
 
+		if state, ok := r.context.MemberState(memberStatus.ID); ok {
+			if state.Reachable {
+				if memberStatus.Conditions.Update(api.ConditionTypeReachable, true, "ArangoDB is reachable", "") {
+					updateMemberStatusNeeded = true
+					nextInterval = nextInterval.ReduceTo(recheckSoonPodInspectorInterval)
+				}
+			} else {
+				if memberStatus.Conditions.Update(api.ConditionTypeReachable, true, "ArangoDB is not reachable", "") {
```
informalict, line 228: true -> false

pkg/deployment/member/state.go
```diff
@@ -0,0 +1,97 @@
...
+
+		if _, err := c.Version(nctx); err != nil {
+			results[id].Reachable = false
+			return
+		}
+
+		results[id].Reachable = true
+	})
+
+	current := map[string]State{}
```
informalict, line 75: what is the idea of `current` if it is local variable and it is not passed anywhere?

informalict (commented):
see comments

ajanikow:
Fixes: https://github.com/arangodb/kube-arangodb/issues/887

PR #924, opened 2022-03-09
[bugfix] Fix golangci-lint tool local installation

nikita-vanyasin:
It was installing globally instead of installing into .gobuild subdir.

The problem was introduced here: https://github.com/arangodb/kube-arangodb/commit/071c837f2bc6111ade71efd17afb457a199659b0#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L394


jwierzbo (approved):

informalict (approved):
LGTM

PR #930, opened 2022-03-16
Multi-arch image discovery

jwierzbo:


pkg/deployment/rotation/arangod.go
```diff
@@ -52,8 +55,24 @@ func affinityCompare(_ api.DeploymentSpec, _ api.ServerGroup, spec, status *core
...
+
+				if spec.Affinity != nil && spec.Affinity.NodeAffinity != nil && spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil {
+					specArchs = api.GetArchsFromNodeSelector(spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms)
+				}
+
+				if status.Affinity != nil && status.Affinity.NodeAffinity != nil && status.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil {
+					statusArchs = api.GetArchsFromNodeSelector(status.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms)
+				}
+
+				if !reflect.DeepEqual(specArchs, statusArchs) {
```
nikita-vanyasin, line 71: Maybe it is possible to move all this new code to a separate function/method? 
e.g.
```
mode = addGracefulRotationIfNeeded(mode, spec, api)
```

pkg/apis/deployment/v1/member_status.go
```diff
@@ -76,6 +76,8 @@ type MemberStatus struct {
 	Image *ImageInfo `json:"image,omitempty"`
 	// OldImage holds old image defails
 	OldImage *ImageInfo `json:"old-image,omitempty"`
+	// Architecture defines Image architecture type
+	Architecture ArangoDeploymentArchitectureType `json:"architecture,omitempty"`
```
ajanikow, line 80: Why ArangoDeploymentArchitectureType instead of *ArangoDeploymentArchitectureType?

It wil lbe easier to determine if arch is set.

pkg/apis/deployment/v1/architecture.go
```diff
@@ -87,3 +73,32 @@ func (a ArangoDeploymentArchitectureType) Validate() error {
...
+			{
+				Key:      k8sutil.NodeArchAffinityLabel,
+				Operator: "In",
+				Values:   []string{string(a)},
+			},
+		},
+	}
+}
+
+func GetArchsFromNodeSelector(selectors []core.NodeSelectorTerm) ArangoDeploymentArchitecture {
```
ajanikow, line 89: Non-api helper function in api? Move to k8sutils

nikita-vanyasin (commented):
please see small improvement suggestion ^

pkg/apis/deployment/v1/architecture.go
ajanikow: + cover case of change from "beta.kubernetes.io/arch" to "kubernetes.io/arch" (it should be still silent)

ajanikow: + run unique on list

ajanikow: + sort them

ajanikow (changes requested):

pkg/apis/deployment/v1/member_status.go
jwierzbo: done

PR #948, opened 2022-04-05
Allow configuration for runAsUser value

nikita-vanyasin:


chart/kube-arangodb/templates/deployment.yaml
```diff
@@ -86,7 +86,7 @@ spec:
             hostIPC: false
             securityContext:
                 runAsNonRoot: true
-                runAsUser: 1000
+                runAsUser: {{ .Values.operator.runAsUser }}
```
ajanikow, line 89: Please, use .Values.operator.securityContext.runAsUser

nikita-vanyasin: done

ajanikow (approved):

PR #956, opened 2022-04-14
[Feature] BackupInProgress & Maintenace Conditions

ajanikow:


pkg/deployment/reconcile/plan_builder_maintenance.go
```diff
@@ -0,0 +1,128 @@
...
+import (
+	"context"
+
+	api "github.com/arangodb/kube-arangodb/pkg/apis/deployment/v1"
+	"github.com/arangodb/kube-arangodb/pkg/util/k8sutil"
+	inspectorInterface "github.com/arangodb/kube-arangodb/pkg/util/k8sutil/inspector"
+	"github.com/rs/zerolog"
+)
+
+func createBackupInProgressConditionPlan(ctx context.Context,
```
jwierzbo, line 32: these two methods are almost the same -> you could use extra arg `ConditionTypeBackup` to parametrise it

PR #962, opened 2022-04-25
[Fix] GT-11 Orphan PVC are not removed

jwierzbo:


pkg/deployment/deployment_inspector.go
```diff
@@ -155,6 +155,7 @@ func (d *Deployment) inspectDeployment(lastInterval util.Interval) util.Interval
 	return nextInterval.ReduceTo(maxInspectionInterval)
 }
 
+// we ensure that the deployment is in a valid state
```
nikita-vanyasin, line 158:
```suggestion
// inspectDeploymentWithError ensures that the deployment is in a valid state
```

pkg/deployment/deployment_inspector.go
```diff
@@ -424,6 +425,7 @@ func (d *Deployment) refreshMaintenanceTTL(ctx context.Context) {
 	}
 }
 
+// here we create all required resources for the deployment
```
nikita-vanyasin, line 428:
```suggestion
// ensureResources creates all required resources for the deployment
```

nikita-vanyasin (approved):
LGMT with a few suggestions

PR #963, opened 2022-04-25
Remove unsafe removal of $ORGDIR from system $GOPATH/src dir

nikita-vanyasin:
The problem:
if you have any changes under `$GOPATH/src/github.com/arangodb`  you will lose all data and history if you execute `make init` in kube-arangodb project.

These commands actually seems to be unnecessary. All generated files are under kube-arangodb directory.
I think the project should not modify external environment in such way.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -237,9 +237,6 @@ update-vendor:
 
 .PHONY: update-generated
 update-generated:
-	@rm -fr $(ORGDIR)
-	@mkdir -p $(ORGDIR)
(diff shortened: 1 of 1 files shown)


jwierzbo (approved):

ajanikow:
@nikita-vanyasin Add if to discover if you are under gopath

nikita-vanyasin:
@ajanikow  Turned out the problem is reproducible only for kube-arangodb-test repository, not the kube-arangodb. Closing

PR #965, opened 2022-04-27
Add operator shutdown handler for graceful termination

nikita-vanyasin:
Not sure how often to check for members termination. Currently set to 1 second. Let me know if adjustment needed.


pkg/operator/operator.go
```diff
@@ -177,8 +180,34 @@ func (o *Operator) Run() {
...
+			time.Sleep(time.Second)
+		}
+		shutdownCh <- struct{}{}
+	}()
+	select {
+	case <-shutdownCh:
+		o.log.Info().Msg("All deployments terminated, exiting.")
+		return
+	case <-time.After(o.Config.ShutdownTimeout):
+		o.log.Info().Msg("Timeout reached, exiting.")
```
informalict, line 208: "Timeout reached, exiting." -> "Timeout reached, all deployments were not terminated in time, exiting."

informalict (approved):
LGTM

pkg/operator/operator.go
nikita-vanyasin: done. Changed to `Timeout reached before all deployments terminated, exiting.`

PR #967, opened 2022-04-28
Fix construction of arangosync member endpoint URLs

nikita-vanyasin:
Some refactorings done to have single point truth for ServerGroup->service port mapping

Only the port is updated, the operator will still use incorrect endpoint in some cases.


ajanikow:
Merged fix with skipped chacks

PR #977, opened 2022-05-04
[Feature] create locals support for actions

informalict:
Allows to add local variables during the specific action.


nikita-vanyasin (approved):
LGTM

PR #984, opened 2022-05-13
[Feature] add agency leader discovery

informalict:


pkg/deployment/agency/cache.go
```diff
@@ -66,48 +69,135 @@ func (c cacheSingle) Data() (State, bool) {
...
+func (c *cache) GetLeaderID() string {
+	return c.leaderID
+}
+
+func (c *cache) Reload(ctx context.Context, clients []agency.Agency) (uint64, error) {
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
-	cfg, err := getAgencyConfig(ctx, client)
+	leaderCli, leaderConfig, err := getLeader(ctx, clients)
```
ajanikow, line 99: If one of Agents is unreachable - then ctx will be already expired. Provide a lot smaller ctx for config discovery

pkg/util/strings.go
```diff
@@ -86,3 +86,14 @@ func DiffStrings(compareWhat, compareTo []string) []string {
 
 	return append(diff, DiffStringsOneWay(compareTo, compareWhat)...)
 }
+
+// IsStringInSlice returns true if haystack contains needle.
```
ajanikow, line 90: USe instead StringListContains

pkg/deployment/agency/cache.go
```diff
@@ -66,48 +69,135 @@ func (c cacheSingle) Data() (State, bool) {
...
-		c.commitIndex = cfg.CommitIndex
-		return cfg.CommitIndex, nil
+		c.commitIndex = leaderConfig.CommitIndex
+		return leaderConfig.CommitIndex, nil
+	}
+}
+
+// getLeader returns config and client to a leader agency.
+// If there is no quorum for the leader then error is returned.
+func getLeader(ctx context.Context, clients []agency.Agency) (agency.Agency, *agencyConfig, error) {
```
ajanikow, line 129: Add additional state if all agents are healthy & all have same commitIndex (so health)

ajanikow (changes requested):

pkg/deployment/agency/cache.go
informalict: done

pkg/util/strings.go
informalict: done

pkg/deployment/agency/cache.go
informalict: done

ajanikow (approved):

PR #991, opened 2022-05-23
[Feature] Add agency leader service

informalict:
test: https://github.com/arangodb/kube-arangodb-test/pull/84

### DIFF
--- a/pkg/deployment/agency/cache.go
+++ b/pkg/deployment/agency/cache.go
@@ -34,7 +34,7 @@ import (
 type health struct {
 	leaderID string
 
-	offsets map[string]uint64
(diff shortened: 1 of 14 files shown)


pkg/deployment/agency/cache.go
```diff
@@ -31,7 +31,15 @@ import (
 	"github.com/arangodb/kube-arangodb/pkg/util/errors"
 )
 
-type health map[string]uint64
+type health struct {
+	leaderID string
+
+	offsets map[string]uint64
```
informalict, line 37: I would change it from `offets` -> `commitIndexes`

PR #1003, opened 2022-06-07
GT-60 OPS CLI with Arango Task

jwierzbo:


pkg/apis/deployment/v1/plan.go
```diff
@@ -231,6 +231,8 @@ type Action struct {
 	Params map[string]string `json:"params,omitempty"`
 	// Locals additional storage for local variables which are produced during the action.
 	Locals PlanLocals `json:"locals,omitempty"`
+	// ID reference of the task involved in this action (if any)
+	TaskID types.UID `json:"taskID,omitempty"`
```
informalict, line 235: please review occurrences of variable `Locals` and add `TaskID` in the same manner.
e.g.:
- `func (a Action) Equal(other Action) bool {`
- `func (in *Action) DeepCopyInto(out *Action) {`

Makefile
```diff
@@ -276,14 +281,17 @@ bin-all: $(BIN) $(VBIN_LINUX_AMD64) $(VBIN_LINUX_ARM64)
 
 $(VBIN_LINUX_AMD64): $(SOURCES) dashboard/assets.go VERSION
 	@mkdir -p $(BINDIR)/$(RELEASE_MODE)/linux/amd64
-	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build ${GOBUILDARGS} --tags "$(RELEASE_MODE)" $(COMPILE_DEBUG_FLAGS) -installsuffix netgo -ldflags "-X $(REPOPATH)/pkg/version.version=$(VERSION) -X $(REPOPATH)/pkg/version.buildDate=$(BUILDTIME) -X $(REPOPATH)/pkg/version.build=$(COMMIT)" -o $(VBIN_LINUX_AMD64) ./
+	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build ${GOBUILDARGS} --tags "$(RELEASE_MODE)" $(COMPILE_DEBUG_FLAGS) -installsuffix netgo -ldflags "-X $(REPOPATH)/pkg/version.version=$(VERSION) -X $(REPOPATH)/pkg/version.buildDate=$(BUILDTIME) -X $(REPOPATH)/pkg/version.build=$(COMMIT)" -o $(VBIN_LINUX_AMD64) ./cmd/main
```
nikita-vanyasin, line 284: If both binaries are expected to have the same build params, it is possible to slightly simplify the code using `$(foreach`

Example in arangosync: https://github.com/arangodb/arangosync/blob/feature/arangosync-migration/Makefile#L157

cmd/cmd_ops.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	goflag "flag"
+
+	"github.com/spf13/cobra"
+	flag "github.com/spf13/pflag"
+)
+
+var (
+	cmdOps = cobra.Command{
+		Use: "arangodb_ops",
```
nikita-vanyasin, line 32: Not sure why it is "arangodb_ops" and not just "ops". With "arangodb_ops" the command would look self-repeatable like `./arangodb_operator_ops arangodb_ops ...`

Makefile
jwierzbo: to implement this more refactor in the project files structure is required since the paths are different (i can not place all main files under the same paths due to the linter requirements)

nikita-vanyasin: @jwierzbo I'm just worried we have another new binary is planned to be added and it would multiply the duplication even more. But I think it can be refactored in a separate PR :+1:

ajanikow (approved):

PR #1007, opened 2022-06-13
[Feature] Unify logging system

ajanikow:


informalict (approved):
TLDR, we should not change logic with a huge refactor, because it would take half a day to review it. I assume that everything is fine so I will give the approval.

ajanikow:
@informalict We did not change logic anywhere - just logger invocation methods

PR #1014, opened 2022-06-15
[Feature] GT-40 Add 'crd install' subcommand

nikita-vanyasin:
The `EnsureCRD` implementation was slightly adjusted to return error if requested, so the command can exit with 1 if the problem occurs.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -289,7 +289,7 @@ func executeMain(cmd *cobra.Command, args []string) {
 			ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 			defer cancel()
 
-			crd.EnsureCRD(ctx, client)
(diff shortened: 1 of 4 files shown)


jwierzbo (approved):

cmd/cmd.go
```diff
@@ -43,6 +43,19 @@ import (
 
 	"github.com/arangodb/kube-arangodb/pkg/deployment/features"
 
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog"
+	"github.com/spf13/cobra"
+	flag "github.com/spf13/pflag"
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
```
ajanikow, line 51: Please, use import names as they are used everywhere

ajanikow (changes requested):

PR #1024, opened 2022-06-20
[Feature] Unify agency access

ajanikow:


pkg/deployment/reconciler/context.go
```diff
@@ -160,7 +160,7 @@ type KubernetesEventGenerator interface {
 }
 
 type DeploymentClient interface {
-	DeploymentAgencyClient
+	//DeploymentAgencyClient
```
nikita-vanyasin, line 163: Probably the comment can be removed

nikita-vanyasin (approved):
LGTM with small suggestion

PR #1036, opened 2022-06-28
[Feature] GT-77 New gRPC and HTTP listeners

nikita-vanyasin:
In future the dashboard and whole pkg/server will be deprecated and removed. The PR creates new listeners and CLI options to control them. The important API endpoints are copied from pkg/server.

See docs/design/api.md for details.


informalict (approved):
LGTM

PR #1038, opened 2022-06-30
Fix `make license` command

nikita-vanyasin:
`addlicense` command expects license header, not the Go code we provided in `make license` command.
As a result, it was adding redundant line comment to license header:
```
// //
// // DISCLAIMER
...
```


informalict (approved):
LGTM

jwierzbo (approved):

PR #1040, opened 2022-06-30
Add docs about integration operator metrics and Prometheus

nikita-vanyasin:
- standard Prometheus installation (no-TLS)
- standard Prometheus installation (TLS)
- Prometheus Operator installation

### DIFF
--- a/docs/design/metrics.md
+++ b/docs/design/metrics.md
@@ -8,6 +8,7 @@ For a full list of available metrics, see [here](./../generated/metrics/README.m
 
 #### Contents
 - [Integration with standard Prometheus installation (no TLS)](#Integration-with-standard-Prometheus-installation-no-TLS)
+- [Integration with standard Prometheus installation (TLS)](#Integration-with-standard-Prometheus-installation-TLS)
 - [Integration with Prometheus Operator](#Integration-with-Prometheus-Operator)
 - [Exposing ArangoDB metrics](#ArangoDB-metrics)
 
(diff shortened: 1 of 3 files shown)


docs/design/metrics.md
```diff
@@ -1,12 +1,59 @@
...
+After creating operator deployment, you must configure Prometheus using a configuration file that instructs it
+about which targets to scrape.
+To do so, add a new scrape job to your prometheus.yaml config:
+```yaml
+scrape_configs:
+  - job_name:       'arangodb-operator'
+
+    scrape_interval: 10s # scrape every 10 seconds.
+
+    scheme: 'https'
```
ajanikow, line 25: When no TLS then why the scheme is defined as HTTPS?

Default Prometheus installation does not support custom CAs, please check `spec.metrics.tls: false` for this.

Also, there are missing annotations used by prometheus.

docs/design/metrics.md
```diff
@@ -1,12 +1,59 @@
...
+    tls_config:
+      insecure_skip_verify: true
+
+    static_configs:
+      - targets:
+          - "<operator-endpoint-ip>:8528"
+```
+
+
+## ArangoDB metrics
```
ajanikow, line 35: There is a collision with PR https://github.com/arangodb/kube-arangodb/pull/1041/files - please, merge it in one if you modify same file

ajanikow (changes requested):

docs/design/metrics.md
nikita-vanyasin: Merged into one PR (this)

docs/design/metrics.md
nikita-vanyasin: @ajanikow 
Current operator server runs only on HTTPS protocol.
If you request with HTTP proto you will get message `Client sent an HTTP request to an HTTPS server.`

`insecure_skip_verify: true` in Prometheus config allows to fetch data with no problems. Tested with latest stable `prom/prometheus` image

nikita-vanyasin: `spec.metrics.tls` applies to ArangoDB sidecard-exporters, which can use HTTP

nikita-vanyasin: Added info on TLS-enabled connection

ajanikow (approved):

PR #1041, opened 2022-06-30
Add docs on setting metrics integration with Prometheus operator

nikita-vanyasin:


examples/metrics/deployment-operator-servicemonitor.yaml
```diff
@@ -11,7 +11,7 @@ metadata:
 spec:
   selector:
     matchLabels:
-      app: arango-deployment-operator
+      app.kubernetes.io/name: kube-arangodb
```
wyattwalter, line 14: This should be noted as a breaking change since it'll not work for those that have adapted to the old hardcoded value.

nikita-vanyasin: Hello. The default labels for operator was changed a long time ago. This change just makes the example up-to-date with current code base. It does not affect the current behaviour of operator.

PR #1045, opened 2022-07-04
GT-80 Memory leaks due to incorrect time.After function usage

jwierzbo:


pkg/deployment/resources/pdbs.go
```diff
@@ -218,10 +218,8 @@ func (r *Resources) ensurePDBForGroup(ctx context.Context, group api.ServerGroup
 		}
 
 		log.Debug("Retry loop for PDB")
-		select {
```
informalict, line 221: variable `action` is not required here

informalict (commented):
2 comments

informalict (approved):
LGTM

PR #1052, opened 2022-07-12
[Feature] get member client from member status

informalict:


pkg/deployment/member/state.go
```diff
@@ -100,44 +111,81 @@ func (s *stateInspector) RefreshState(ctx context.Context, members api.Deploymen
...
+		cs.NotReachableErr = globals.GetGlobalTimeouts().ArangoDCheck().RunWithTimeout(ctx, func(ctxChild context.Context) error {
+			v, err := cli.Version(ctxChild)
+			if err == nil {
+				cs.Version = v
+			}
+			return err
+		})
+
+		if s.deployment.GetMode() == api.DeploymentModeCluster {
+			// Fetch health only in cluster mode.
```
ajanikow, line 145: In all modes - we should never connect via service

informalict: fixed

ajanikow (approved):

Issue #1068, opened 2022-07-22
Unable to deploy operator along with ArangoDeployment CRD


billythekidpl:
Hi everyone. Lately I've been trying to update my kube-arangodb from version `1.1.5` to `1.2.15` but I've enforced a problem. It seems that with that version, I'm unable to deploy arango-operator along with ArangoDeployment object. Everything works fine when I don't define ArangoDeployment. 

POD logs are as follows:
```
I0722 14:35:33.727108       1 cmd.go:246] nice to meet you
2022-07-22T14:35:33Z INF Starting arangodb-operator (Community), version 1.2.15 build e9c43416 operator-id=9ljmm pod-name=arango-xxx-operator-74d4484bfc-9ljmm pod-namespace=xxx-9211-bump-arango-eks
2022-07-22T14:35:33Z INF Serving on 0.0.0.0:8528 operator-id=9ljmm
I0722 14:35:33.841831       1 leaderelection.go:243] attempting to acquire leader lease xxx-9211-bump-arango-eks/arango-deployment-operator...
I0722 14:35:33.947019       1 leaderelection.go:253] successfully acquired lease xxx-9211-bump-arango-eks/arango-deployment-operator
2022-07-22T14:35:33Z INF New leader detected identity=arango-xxx-operator-74d4484bfc-9ljmm-arango-xxx-operator-74d4484bfc-9ljmm lock-name=arango-deployment-operator namespace=xxx-9211-bump-arango-eks operator-id=9ljmm
2022-07-22T14:35:33Z INF Event(v1.ObjectReference{Kind:"Endpoints", Namespace:"xxx-9211-bump-arango-eks", Name:"arango-deployment-operator", UID:"7f0524b8-af7e-4a1c-b2f2-00e5238e784c", APIVersion:"v1", ResourceVersion:"198934152", FieldPath:""}): type: 'Normal' reason: 'LeaderElection' arango-xxx-operator-74d4484bfc-9ljmm-arango-xxx-operator-74d4484bfc-9ljmm became leader operator-id=9ljmm
[...]

PR #1074, opened 2022-07-26
[Bugfix] GT-131 Ensure that client cache is initialized before using it

jwierzbo:


ajanikow:
Fixes: https://github.com/arangodb/kube-arangodb/issues/1068

PR #1081, opened 2022-08-08
[Generated] Add Timezones

ajanikow:


jwierzbo (approved):

Makefile
```diff
@@ -532,7 +538,7 @@ check-community:
 _check:
 	@$(MAKE) fmt license-verify linter run-unit-tests bin
 
-generate-documentation: generate-go-documentation fmt
+generate-documentation: generate-go-documentation fmt-generated
```
nikita-vanyasin, line 541: Probably it's time to rename this target to 'generate' or something like that because it's related not only to docs now

Makefile
```diff
@@ -191,7 +191,13 @@ license-verify:
 fmt:
 	@echo ">> Ensuring style of files"
 	@$(GOPATH)/bin/goimports -w $(SOURCES)
-	@$(GOPATH)/bin/gci write -s "standard" -s "default" -s "prefix(github.com/arangodb)" -s "prefix(github.com/arangodb/kube-arangodb)" $(SOURCES)  
+	@$(GOPATH)/bin/gci write -s "standard" -s "default" -s "prefix(github.com/arangodb)" -s "prefix(github.com/arangodb/kube-arangodb)" $(SOURCES) 
+
+.PHONY: fmt-generated
+fmt-generated:
```
nikita-vanyasin, line 197: Why not format these files in same 'fmt' target?

Also, why not gormat all files under '$(ROOT)/pkg/generated/' ?
Adding specific directories makes makefile aware of too many things

nikita-vanyasin (commented):
Added a few comments on Makefile changes

nikita-vanyasin (approved):

PR #1088, opened 2022-08-17
[Feature] Timezone

ajanikow:


pkg/deployment/rotation/arangod_volumes.go
```diff
@@ -0,0 +1,190 @@
...
+		specV := mapVolumeMounts(spec)
+		statusV := mapVolumeMounts(status)
+
+		diff := getVolumeMountsDiffFromPods(specV, statusV)
+
+		if len(diff) == 0 {
+			return SkippedRotation, nil, nil
+		}
+
+		for k, v := range diff {
```
nikita-vanyasin, line 123: There is exactly same loop in `comparePodVolumes` func, maybe some common function can be introduced?

ajanikow: One is volumeMount, second volume - so it will be hard to do it in this way,

pkg/apis/deployment/v1/plan.go
```diff
@@ -199,7 +199,8 @@ const (
 	ActionTypeRebalancerClean    ActionType = "RebalancerClean"
 
 	// Resources
-	ActionTypeResourceSync ActionType = "ResourceSync"
+	ActionTypeResourceSync  ActionType = "ResourceSync"
+	ActionTypeTimezoneCMSet ActionType = "TimezoneCMSet"
```
nikita-vanyasin, line 203: What is CM abbreviation in name?

nikita-vanyasin (approved):

PR #1105, opened 2022-09-13
Workaround for kustomization

cristiansteib:
The original changes do not work.


jwierzbo:
Fixed in: https://github.com/arangodb/kube-arangodb/commit/58a3789fcc2092f5a608e85a3c1d78e38d4db013

PR #1134, opened 2022-09-28
ensure consistency during cancellation of replication

informalict:


pkg/apis/replication/v1/replication_status.go
```diff
@@ -36,7 +36,7 @@ type DeploymentReplicationStatus struct {
 	// Destination contains the detailed status of the destination endpoint
 	Destination EndpointStatus `json:"destination"`
 
-	// CancelFailures records the number of times that the configuration was canceled
+	// Deprecated: CancelFailures records the number of times that the configuration was canceled
```
nikita-vanyasin, line 39:
```suggestion
	// Deprecated: this field will not be updated anymore
	// CancelFailures records the number of times that the configuration was canceled
```

pkg/apis/replication/v1/conditions.go
```diff
@@ -31,6 +31,17 @@ type ConditionType string
 const (
 	// ConditionTypeConfigured indicates that the replication has been configured.
 	ConditionTypeConfigured ConditionType = "Configured"
+	// ConditionTypeEnsuredInSync indicates that the replication consistency was checked.
+	ConditionTypeEnsuredInSync ConditionType = "EnsuredInSync"
+	// ConditionTypeAborted indicates that the replication was canceled with abort option.
+	ConditionTypeAborted ConditionType = "Aborted"
+
+	// ConditionTypeConfiguredActive describes synchronization as active.
+	ConditionTypeConfiguredActive = "Active"
```
nikita-vanyasin, line 40: I think these constants (active, inactive, invalid) should named "ConditionReasonXXXYYY".
Because currently it might be confused with ConditionType

pkg/replication/finalizers.go
```diff
@@ -187,3 +255,81 @@ func removeDeploymentReplicationFinalizers(crcli versioned.Interface, p *api.Ara
 	}
 	return nil
 }
+
+// isFinalizerExist returns true if a given finalizer exists.
+func isFinalizerExist(p *api.ArangoDeploymentReplication, finalizer string) bool {
```
nikita-vanyasin, line 260:
```suggestion
func finalizerExists(p *api.ArangoDeploymentReplication, finalizer string) bool {
```

nikita-vanyasin (approved):
LGTM with few suggestions

pkg/apis/replication/v1/conditions.go
informalict: changed

pkg/replication/finalizers.go
informalict: changed

ajanikow (approved):

PR #1149, opened 2022-10-19
[Bugfix] Fix Go routine leaks

ajanikow:


pkg/util/timer/after_test.go
```diff
@@ -23,8 +23,31 @@ package timer
...
+		<-After(duration)
+	}
+}
+
+func Test_MemoryLeak_After(t *testing.T) {
+	t.Run("Initial wrap", func(t *testing.T) {
+		afterManyTimes(time.Nanosecond, 1024^2)
+	})
+
+	t.Run("Memory should me stable", func(t *testing.T) {
```
nikita-vanyasin, line 45:
```suggestion
	t.Run("Memory should be stable", func(t *testing.T) {
```

pkg/util/timer/after_test.go
```diff
@@ -23,8 +23,31 @@ package timer
...
+		start := memory.Sys()
+		for i := 0; i < 256; i++ {
+			afterManyTimes(time.Nanosecond, 1024^2)
+			now := memory.Sys()
+			t.Logf("Memory grow by %d", now-start)
```
nikita-vanyasin, line 50: Maybe test should fail if memory has grown more than expected?

ajanikow: We cant predict it - it is used to check via logs if there is grow. I can remove this test due to fact it is fixed

nikita-vanyasin (approved):

jwierzbo (approved):

PR #1150, opened 2022-10-19
GT-196 Update DeploymentReplicationStatus if configuring synchronisation errored

nikita-vanyasin:
if user provided wrong config or source endpoint is unreachable / error-ed, we should report it on replication status CR


jwierzbo (approved):

informalict (approved):
LGTM

ajanikow (approved):

Issue #1143, opened 2022-10-13
Arango ID Pod cannot be created


4gsim:
When running on an kubernetes cluster with the pod security level set on restricted the arango ID pod cannot be created.

The error is:
DBG Failed to create image ID pod error="pods \"arango-id-41ae7c\" is forbidden: violates PodSecurity \"restricted:latest\": seccompProfile (pod or container \"server\" must set securityContext.seccompProfile.type to \"RuntimeDefault\" or \"Localhost\")" image=arangodb/enterprise:3.7.15 name=arango namespace=playground operator-id=pvdgq pod=arango-id-41ae7c

You cannot set this seccompProfile on the id pod (or any other arango pod for that matter) in the arango deployment manifest.

PR #1151, opened 2022-10-20
[Feature] Extend Pod Security context

ajanikow:
Fixes: https://github.com/arangodb/kube-arangodb/issues/1143


pkg/apis/deployment/v1/server_group_security_context_spec.go
```diff
@@ -0,0 +1,125 @@
...
+	return *s.DropAllCapabilities
+}
+
+// GetAddCapabilities add capabilities to pod context
+func (s *ServerGroupSpecSecurityContext) GetAddCapabilities() []core.Capability {
+	if s == nil {
+		return nil
+	}
+
+	if s.AddCapabilities == nil {
```
informalict, line 69: we can remove these 3 lines, because if it is nil then it will be returned at the end of the function

pkg/apis/deployment/v2alpha1/server_group_security_context_spec.go
```diff
@@ -0,0 +1,125 @@
...
+	return *s.DropAllCapabilities
+}
+
+// GetAddCapabilities add capabilities to pod context
+func (s *ServerGroupSpecSecurityContext) GetAddCapabilities() []core.Capability {
+	if s == nil {
+		return nil
+	}
+
+	if s.AddCapabilities == nil {
```
informalict, line 69: the same here

informalict (approved):
LGTM with small comment

PR #1161, opened 2022-11-01
GT-137 image validation for AMD64 arch

jwierzbo:
This validation handles two cases:
1. `arch` annotation - we will not let the change arch in member if ArangoDB image is below 3.10
2. member creation - we will change the member arch to the AMD64 if ArangoDB image is below 3.10

### DIFF
--- a/pkg/apis/deployment/v1/conditions.go
+++ b/pkg/apis/deployment/v1/conditions.go
@@ -78,6 +78,8 @@ const (
 	ConditionTypeUpgradeFailed ConditionType = "UpgradeFailed"
 	// ConditionTypeArchitectureMismatch indicates that the member has a different architecture than the deployment.
 	ConditionTypeArchitectureMismatch ConditionType = "ArchitectureMismatch"
+	// ConditionTypeArchitectureARM64CannotBeApplied indicates that the member has a different architecture than the requested one.
+	ConditionTypeArchitectureARM64CannotBeApplied ConditionType = "ArchitectureARM64CannotBeApplied"
 
 	// ConditionTypeMemberMaintenanceMode indicates that Maintenance is enabled on particular member
 	ConditionTypeMemberMaintenanceMode ConditionType = "MemberMaintenanceMode"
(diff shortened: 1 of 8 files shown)


pkg/util/k8sutil/events.go
```diff
@@ -283,3 +283,12 @@ func NewOperatorEngineOpsAlertEvent(reason string, apiObject APIObject) *Event {
 	event.Message = fmt.Sprintf("Event OperatorEngineOpsAlert raised, investigation needed: %s", reason)
 	return event
 }
+
+// NewCannotSetArchitectureARM64Event creates an even of type CannotSetArchitectureARM64.
+func NewCannotSetArchitectureARM64Event(apiObject runtime.Object, memberId string) *Event {
+	event := newDeploymentEvent(apiObject)
+	event.Type = core.EventTypeWarning
+	event.Reason = "Can not set architecture ARM64"
+	event.Message = fmt.Sprintf("Can not apply ARM64 arch for member %s. It is not supported for ArangoDB < 3.10.0", memberId)
```
nikita-vanyasin, line 292: The function named like a general "can't set architecture event" but actually it reports the reason: ArangoD version incompatibility. Maybe pass the message or message part through func parameter?

nikita-vanyasin (approved):

pkg/apis/deployment/v1/conditions.go
```diff
@@ -78,6 +78,8 @@ const (
 	ConditionTypeUpgradeFailed ConditionType = "UpgradeFailed"
 	// ConditionTypeArchitectureMismatch indicates that the member has a different architecture than the deployment.
 	ConditionTypeArchitectureMismatch ConditionType = "ArchitectureMismatch"
+	// ConditionTypeArchitectureARM64CannotBeApplied indicates that the member has a different architecture than the requested one.
+	ConditionTypeArchitectureARM64CannotBeApplied ConditionType = "ArchitectureARM64CannotBeApplied"
```
ajanikow, line 82: Add one condition for all archs + reason

pkg/util/k8sutil/events.go
ajanikow: +1

ajanikow (changes requested):

PR #1165, opened 2022-11-02
[Feature] Agency ArangoSync State check

ajanikow:


nikita-vanyasin (approved):

pkg/deployment/agency/arangosync_test.go
```diff
@@ -0,0 +1,47 @@
...
+	require.True(t, s.Agency.ArangoDB.ArangoSync.ArangoSync.State.Outgoing.Targets.Exists())
+	require.True(t, s.Agency.ArangoDB.ArangoSync.IsSyncInProgress())
+}
+
+func Test_Sync_Source(t *testing.T) {
+	var s DumpState
+	require.NoError(t, json.Unmarshal(syncTarget, &s))
+
+	require.NotNil(t, s.Agency.ArangoDB.ArangoSync.ArangoSync)
+	require.NotNil(t, s.Agency.ArangoDB.ArangoSync.ArangoSync.State.Incoming.State)
```
informalict, line 44: why do we have here ArangoSync twice?

pkg/deployment/agency/state.go
```diff
@@ -87,13 +87,18 @@ func (c *cache) loadState(ctx context.Context, client agency.Agency) (State, err
...
+	Arango   State   `json:"arango"`
+	ArangoDB StateDB `json:"arangodb,omitempty"`
 }
 
 type DumpState struct {
 	Agency StateRoot `json:"agency"`
 }
 
+type StateDB struct {
+	ArangoSync ArangoSyncLazy `json:"arangoSync,omitempty"`
```
informalict, line 99:
```suggestion
	ArangoSync ArangoSyncLazy `json:"arangosync,omitempty"`
```

informalict (commented):
comments

pkg/deployment/agency/arangosync_test.go
ajanikow: It is due to Lazy loader - in case of failure I do not want to break Operator Deployment loop

informalict (approved):
LGTM

PR #1172, opened 2022-11-03
Don't serialize and deprecate two DeploymentReplicationStatus fields

nikita-vanyasin:
These two fields might be too big - exceeding k8s limits. e.g. one of customers has 10k shards - there will be at least about `10000*2*50=1MB` of data


informalict (approved):
LGTM

PR #1179, opened 2022-11-11
Do not check checksums for DeploymentReplicationStatus.IncomingSynchronization field values

nikita-vanyasin:


informalict (approved):
LGTM

PR #1188, opened 2022-11-21
[Bugfix] Quick checking ArangoSync availability

informalict:
During tests I found that ArangoSync master was not UP for long time, because replication was in initial-sync. In case of OpSec it takes 2 hours to change this state into running. Operator waits for running state and after 30 minutes gives up and then it starts next ArangoSync master. If CPU, memory resources are not sufficient then master is killed before 30 minutes, and replication never can reach running state. 

I have changed `Health` endpoint into `Version` thanks to that ArangoSync is immediately responsive.


pkg/deployment/reconcile/action_wait_for_member_up.go
```diff
@@ -197,7 +197,10 @@ func (a *actionWaitForMemberUp) checkProgressArangoSync(ctx context.Context) (bo
 		a.log.Err(err).Debug("Failed to create arangosync client")
 		return false, false, nil
 	}
-	if err := c.Health(ctx); err != nil {
+
+	// When replication is in initial-sync state, then it can take a long time to be in running state.
+	// This is the reason why Health of ArangoSync can not be checked here.
+	if _, err := c.Version(ctx); err != nil {
 		a.log.Err(err).Debug("Health not ok yet")
```
nikita-vanyasin, line 204:
```suggestion
		a.log.Err(err).Debug("Version not ok yet")
```
or
```suggestion
		a.log.Err(err).Debug("Member is not ready yet")
```

nikita-vanyasin (approved):

jwierzbo (approved):

PR #1203, opened 2022-12-08
[Bugfix] Do not stop Sync if Synchronization is in progress

ajanikow:
Manual tests went fine:
- ArangoDeploymentReplication was removed so agency was updated to inactive sync.
- service deployment-sync was removed.
- as a last step ArangoSync pods were removed.


informalict (approved):
LGTM

PR #1222, opened 2022-12-30
[Bugfix] Switch tolerations compare

ajanikow:


pkg/deployment/reconcile/action_runtime_sync_tolerations.go
```diff
@@ -89,6 +91,8 @@ func (a actionRuntimeContainerSyncTolerations) Start(ctx context.Context) (bool,
 		return false, errors.Wrapf(err, "Unable to create patch")
 	}
 
+	println("XXXXXX", string(p))
```
informalict, line 94: is it intentional?

informalict (approved):
small comment

pkg/deployment/reconcile/action_runtime_sync_tolerations.go
ajanikow: Fixed already

PR #1237, opened 2023-01-26
[Feature] CreationFailed ArangoMember Phase

ajanikow:


pkg/deployment/resources/pod_inspector.go
```diff
@@ -471,7 +471,7 @@ func (r *Resources) InspectPods(ctx context.Context, cachedStatus inspectorInter
 			if _, exists := cachedStatus.Pod().V1().GetSimple(podName); !exists {
 				log.Str("pod-name", podName).Debug("Does not exist")
 				switch m.Phase {
-				case api.MemberPhaseNone, api.MemberPhasePending:
+				case api.MemberPhaseNone, api.MemberPhasePending, api.MemberPhaseCreationFailed:
 					// Do nothing
 					log.Str("pod-name", podName).Debug("PodPhase is None, waiting for the pod to be recreated")
```
informalict, line 476: We have to change message: "PodPhase is None, waiting for the pod to be recreated" -> "PodPhase is %s, waiting for the pod to be recreated" where %s is a phase

informalict (approved):
One little comment

pkg/deployment/resources/pod_inspector.go
ajanikow: @informalict Above statement (in code) is valid. All of this states represent PodPhase (which is just none - pod is not created - for corresponding MemberPhases)

PR #1239, opened 2023-02-06
[Improvement] GT-228 Add new field to CR for more precise calculation of DC2DC replication progress

nikita-vanyasin:
Deprecate `ShardsTotal` and `ShardsInSync`  fields. Clients should use new `Progress` field directly instead of calculating it. (also ShardsTotal might have incorrect value when starting replication of huge dataset)


jwierzbo (approved):

informalict (approved):
lgtm

PR #1243, opened 2023-02-16
GT-341 Fix invalid Timeout calculation in case of ActionList

jwierzbo:


pkg/deployment/reconcile/plan_executor.go
```diff
@@ -396,18 +396,30 @@ func (d *Reconciler) executeAction(ctx context.Context, planAction api.Action, a
...
+	if planAction.StartTime == nil {
+		return false
+	}
+	if planAction.StartTime.IsZero() {
+		return false
+	}
+	if timeout.Infinite() {
+		return false
+	}
+	return time.Now().After(planAction.StartTime.Add(timeout.Duration))
```
informalict, line 420: you can use `time.Since(planAction.StartTime) > timeout.Duration`

informalict (approved):
LGTM with small suggestion

pkg/deployment/reconcile/plan_executor.go
ajanikow: 👍

PR #1246, opened 2023-02-20
[Feature] Optional ResignLeadership

ajanikow:
After: https://github.com/arangodb/kube-arangodb/pull/1245


pkg/apis/deployment/v1/actions.generated.go
```diff
@@ -759,7 +759,7 @@ func (a ActionType) Optional() bool {
 	case ActionTypeRenewTLSCertificate:
 		return false
 	case ActionTypeResignLeadership:
-		return false
+		return true
```
informalict, line 762: it could be generated in a way where only optional actions returns true and default returns false, so it would be more legible.

ajanikow: I will do it in PR if it is fine

PR #1261, opened 2023-03-01
GT-198 Cleanout calculation - picks members with the lowest number of shards

jwierzbo:


pkg/deployment/agency/state.go
```diff
@@ -132,6 +132,42 @@ func (s State) CountShards() int {
 	return count
 }
 
+// ShardsByDbServers returns a map of DBServers and the amount of shards they have
+func (s State) ShardsByDbServers() map[Server]int {
```
nikita-vanyasin, line 136:
```suggestion
func (s State) ShardsByDBServers() map[Server]int {
```

nikita-vanyasin (approved):

informalict (approved):
lgtm

pkg/apis/deployment/v1/member_status.go
```diff
@@ -53,7 +53,7 @@ type MemberStatus struct {
 	CreatedAt meta.Time `json:"created-at"`
 	// Conditions specific to this member
 	Conditions ConditionList `json:"conditions,omitempty"`
-	// RecentTerminatons holds the times when this member was recently terminated.
+	// RecentTerminations holds the times when this member was recently terminated.
```
ajanikow, line 56: Synchronize versions

ajanikow (changes requested):

ajanikow (approved):

PR #1269, opened 2023-03-08
GT-143 ArangoBackup: create retries and MaxIterations limit

jwierzbo:


pkg/apis/backup/v1/zz_generated.deepcopy.go
```diff
@@ -310,6 +310,11 @@ func (in *ArangoBackupSpecBackOff) DeepCopyInto(out *ArangoBackupSpecBackOff) {
 		*out = new(int)
 		**out = **in
 	}
+	if in.MaxIterations != nil {
```
informalict, line 313: it should be also in v1alpha

jwierzbo: we do not have v1alpha for Backups

ajanikow (approved):

PR #1276, opened 2023-03-21
[Feature] Disable Pod termination requirement

ajanikow:


pkg/deployment/reconcile/action_drain_member.go
```diff
@@ -0,0 +1,108 @@
...
+// actionDrainMember implements an DrainMemberAction.
+type actionDrainMember struct {
+	// actionImpl implement timeout and member id functions
+	actionImpl
+}
+
+// Start performs the start of the action.
+// Returns true if the action is completely finished, false in case
+// the start time needs to be recorded and a ready condition needs to be checked.
+func (a *actionDrainMember) Start(ctx context.Context) (bool, error) {
```
jwierzbo, line 49: Please add a short description of that Action:
- what is the purpose of the action
- what is the cause of adding it

jwierzbo (changes requested):

PR #1292, opened 2023-04-13
GT-403 Deprecate Endpoint field in ArangoDeployment

jwierzbo:


informalict (changes requested):
run `make fix` on this branch, so it will tweak headers

pkg/apis/deployment/v1/member_status.go
```diff
@@ -73,8 +72,7 @@ type MemberStatus struct {
 	Architecture *ArangoDeploymentArchitectureType `json:"architecture,omitempty"`
 	// Upgrade define if upgrade should be enforced during next execution
 	Upgrade bool `json:"upgrade,omitempty"`
-	// Endpoint definition how member should be reachable
-	Endpoint *string `json:"endpoint,omitempty"`
```
ajanikow, line 77: Just mark as deprecated, do not remove

ajanikow (changes requested):

pkg/apis/deployment/v1/member_status.go
jwierzbo: It is not removed - it's moved the the deprecated section below....

PR #1295, opened 2023-04-20
[Feature] ]Add fast restart annotation

informalict:
tests: https://github.com/arangodb/kube-arangodb-test/pull/169

### DIFF
--- a/pkg/apis/deployment/annotations.go
+++ b/pkg/apis/deployment/annotations.go
@@ -1,7 +1,7 @@
 //
 // DISCLAIMER
 //
-// Copyright 2016-2022 ArangoDB GmbH, Cologne, Germany
+// Copyright 2016-2023 ArangoDB GmbH, Cologne, Germany
 //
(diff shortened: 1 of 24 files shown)


pkg/apis/deployment/annotations.go
```diff
@@ -26,6 +26,7 @@ const (
 	ArangoDeploymentPodChangeArchAnnotation         = ArangoDeploymentAnnotationPrefix + "/arch"
 	ArangoDeploymentPodRotateAnnotation             = ArangoDeploymentAnnotationPrefix + "/rotate"
 	ArangoDeploymentPodReplaceAnnotation            = ArangoDeploymentAnnotationPrefix + "/replace"
+	ArangoDeploymentPodFastRestart                  = ArangoDeploymentAnnotationPrefix + "/fast_restart"
```
jwierzbo, line 29: can we add here some short description how it works?

informalict: actually it is draft PR, but yes I think we should have one place where all annotations should be described.

PR #1296, opened 2023-04-23
GT-405 Automatically fix out of sync shards with broken MerkleTree

jwierzbo:


pkg/deployment/reconcile/plan_builder_rebuild_outsynced_shards.go
```diff
@@ -0,0 +1,121 @@
...
+			shouldUpdate := false
+			if _, ok := outSyncedMembers[member.ID]; ok {
+				shouldUpdate = member.Conditions.Update(api.ConditionTypeOutSyncedShards, true, "Member has out-synced shard(s)", "")
+			} else {
+				shouldUpdate = member.Conditions.Remove(api.ConditionTypeOutSyncedShards)
+			}
+
+			if shouldUpdate {
+				if err := context.UpdateMember(ctx, member); err != nil {
+					r.log.Error("Can not save member condition", member.ID, api.ConditionTypeOutSyncedShards, err)
```
informalict, line 109: lets say err is nil here, so member is not updated, but below variable `disableRebuildOutSyncedShards` will  be set to true, and this action will not be launched. Is it intended?

jwierzbo: Right now yes - we are not using that condition anywhere

docs/design/rebuild_out_synced_shards.md
```diff
@@ -0,0 +1,17 @@
+# Force rebuild out-synced Shards with broken Merkle Tree
+
+## Overview
+
+TODO
```
nikita-vanyasin, line 5: Adding this will help understanding the code in this PR

docs/design/rebuild_out_synced_shards.md
```diff
@@ -0,0 +1,17 @@
...
+
+## How to use
```
nikita-vanyasin, line 7: on main README it is mentioned that 'user is aware of the risks'. Please describe these risks here.

pkg/deployment/features/rebuild_out_synced_shards.go
```diff
@@ -0,0 +1,38 @@
...
+
+func init() {
+	registerFeature(rebuildOutSyncedShards)
+}
+
+var rebuildOutSyncedShards = &feature{
+	name:               "force-rebuild-out-synced-shards",
+	description:        "Force rebuild permanently out-synced shards due to a bug in ArangoDB 3.8-3.10",
+	version:            "3.8.0",
+	enterpriseRequired: true,
```
nikita-vanyasin, line 31: Inconsistency: in README it says that this feature can be applied to community version too.

docs/design/rebuild_out_synced_shards.md
jwierzbo: due to lack of time i will add it as a separate PR

docs/design/rebuild_out_synced_shards.md
jwierzbo: will add in the next PR with updated doc

PR #1305, opened 2023-05-02
Fix 'make generate' command in ZSH

jwierzbo:


nikita-vanyasin (commented):
Not sure how pv_creator.go is related to makefile adjustments, but otherwise LGTM

ajanikow (approved):

PR #1314, opened 2023-05-24
Fix creating sync components with EA type set to Managed and headless svc

nikita-vanyasin:
If customer uses "Type=Managed" for `SyncExternalAccessSpec` AND uses headless service (`ClusterIP=None`) then sync pods won't be created due to an error while creating a HostAlias for pod.

The code in `syncHostAlias` method still not perfect because it assumes that EA service name would be always `<depl-name>-sync`. That should be fixed in separate ticket.


informalict (approved):
lgtm

PR #1318, opened 2023-05-30
GT-428 Add disallowConcurrent option to ArangoBackupPolicy

nikita-vanyasin:
If true, ArangoBackups won't be created if previous backup still running (per policy).


informalict (approved):
lgtm

PR #1324, opened 2023-06-05
[Maintenance] GT-435 | Update ArangoDB version used in examples YAMLs

nikita-vanyasin:
set to 3.10.6 because 3.11.0 has some known issues. In most cases it is better to use 3.10 branch until 3.11.1 is not released.


jwierzbo (approved):

informalict (approved):
lgtm

PR #1421, opened 2023-09-26
GT-478 Migration to CircleCI

jwierzbo:


nikita-vanyasin (commented):
@jwierzbo I think you can drop travisci in same PR

nikita-vanyasin (approved):

PR #1468, opened 2023-10-29
(Feature) Add validation rules for CRDs

nikita-vanyasin:


internal/cr_validation_test.go
```diff
@@ -0,0 +1,393 @@
...
+
+	if len(docName) > 0 {
+		s.ExternalDocs = &apiextensions.ExternalDocumentation{
+			URL: getDocsLinkForField(docName, def),
+		}
+	}
+	s.Description = strings.Join(def.Docs, "\n")
+}
+
+func Test_GenerateCRValidationSchemas(t *testing.T) {
```
jwierzbo, line 70: We should add a comment here that we used that test to generate schema files

nikita-vanyasin: done

jwierzbo (dismissed):

jwierzbo (changes requested):
Please add doc for that feature under doc/features

nikita-vanyasin:
> Please add doc for that feature under doc/features

Done. See `crds.md`
I've added it to docs root because `feature` contains only features generated from features.yaml - which is not the case for CRD validation.

jwierzbo (approved):

pkg/crd/crds/backups-backup.schema.generated.json
```diff
@@ -0,0 +1,208 @@
+{
```
ajanikow, line 1: Try to parse it to yaml - use k8s.io/apimachinery/pkg/util/yaml

pkg/crd/crds/backups-backuppolicy.go
```diff
@@ -34,23 +33,26 @@ const (
...
-	if err := yaml.Unmarshal(backupsBackupPolicy, &backupsBackupPolicyCRD); err != nil {
-		panic(err)
-	}
+	mustLoadCRD(backupsBackupPolicy, backupsBackupPolicySchemaRaw, &backupsBackupPolicyCRD, &backupsBackupPolicyCRDWithSchema)
 }
 
-func BackupsBackupPolicyPolicy() *apiextensions.CustomResourceDefinition {
-	return backupsBackupPolicyCRD.DeepCopy()
+func BackupsBackupPolicyPolicy(opts ...GetCRDOptions) *apiextensions.CustomResourceDefinition {
+	return getCRD(backupsBackupPolicyCRD, backupsBackupPolicyCRDWithSchema, opts...)
```
ajanikow, line 40: Just unmarshal it twice instead of creating new func

ajanikow: BackupsBackupPolicyPolicy -> this should have same working function, add BackupsBackupPolicyPolicyWithOptions

ajanikow: BackupsBackupPolicyPolicyWithOptions should default options to the one which is used by CRD -> for ML True, others False (if not set at all)

ajanikow (changes requested):

pkg/crd/crds/backups-backup.schema.generated.json
nikita-vanyasin: Done

nikita-vanyasin:
Will split PR into two smaller

PR #1470, opened 2023-11-02
GT-526 Use agency cache lock in metrics exporter

jwierzbo:


pkg/deployment/agency/cache.go
```diff
@@ -144,7 +144,11 @@ type Health interface {
 
 type Cache interface {
 	Reload(ctx context.Context, size int, clients Connections) (uint64, error)
+	// Deprecated: Use Apply instead.
+	// It can cause to Read/Write error when state is reloaded.
```
nikita-vanyasin, line 148:
```suggestion
	// It can cause Read/Write error when state is reloaded.
```

pkg/deployment/agency/cache.go
```diff
@@ -144,7 +144,11 @@ type Health interface {
...
 	Data() (state.State, bool)
+	// Apply applies a function to the current state.
```
nikita-vanyasin, line 150:
```suggestion
	// Apply applies a function to the current state. Returns true if function was applied
```

pkg/deployment/reconciler/context.go
```diff
@@ -96,6 +96,11 @@ type DeploymentImageManager interface {
 }
 
 type ArangoAgencyGet interface {
+	// WithAgencyCache executes the given action with the agency cache using a Read lock.
```
nikita-vanyasin, line 99:
```suggestion
	// WithAgencyCache executes the given action with the agency cache using a Read lock. Returns true if action was applied
```

nikita-vanyasin (approved):

PR #1471, opened 2023-11-02
GT-509 Remove scale_down_candidate annotation

jwierzbo:


pkg/apis/deployment/v1/conditions.go
```diff
@@ -76,8 +76,6 @@ const (
 	ConditionTypeMemberVolumeUnschedulable ConditionType = "MemberVolumeUnschedulable"
 	// ConditionTypeMarkedToRemove indicates that the member is marked to be removed.
 	ConditionTypeMarkedToRemove ConditionType = "MarkedToRemove"
-	// ConditionTypeScaleDownCandidate indicates that the member will be picked in ScaleDown operaion.
```
nikita-vanyasin, line 79: That might be a breaking change for some customer who is using these APIs. I thought we can't do such changes in patch release?

ajanikow: True, it should not be removed - just deprecated

ajanikow (changes requested):
Revert all changes in public api, you can remove only internal parts

pkg/apis/deployment/v1/conditions.go
```diff
@@ -76,7 +76,8 @@ const (
...
+	// ConditionTypeScaleDownCandidate indicates that the member will be picked in ScaleDown operation (Currently not used).
+	// @deprecated will be removed in 1.3.0
```
nikita-vanyasin, line 80:
```suggestion
	// Deprecated: will be removed in 1.3.0
```

pkg/apis/deployment/v2alpha1/conditions.go
```diff
@@ -76,7 +76,8 @@ const (
 	ConditionTypeMemberVolumeUnschedulable ConditionType = "MemberVolumeUnschedulable"
 	// ConditionTypeMarkedToRemove indicates that the member is marked to be removed.
 	ConditionTypeMarkedToRemove ConditionType = "MarkedToRemove"
-	// ConditionTypeScaleDownCandidate indicates that the member will be picked in ScaleDown operaion.
+	// ConditionTypeScaleDownCandidate indicates that the member will be picked in ScaleDown operation (Currently not used).
+	// @deprecated will be removed in 1.3.0
```
nikita-vanyasin, line 80:
```suggestion
	// Deprecated: will be removed in 1.3.0
```

nikita-vanyasin (approved):

ajanikow (approved):

PR #1480, opened 2023-11-08
[Feature] Compare Generic

ajanikow:


pkg/util/compare/compare.go
```diff
@@ -0,0 +1,95 @@
...
+//
+
+package compare
+
+import (
+	api "github.com/arangodb/kube-arangodb/pkg/apis/deployment/v1"
+	"github.com/arangodb/kube-arangodb/pkg/logging"
+)
+
+func P2[T interface{}, P1, P2 interface{}](
```
nikita-vanyasin, line 28: Is it possible to give this function a name?
If not, please add the comment what it does

pkg/util/compare/compare.go
```diff
@@ -0,0 +1,95 @@
...
+			if specData != "" {
+				line = line.Str("spec", specData)
+			}
+
+			if statusData != "" {
+				line = line.Str("status", statusData)
+			}
+		}
+
+		line.Info("Pod needs rotation - templates does not match")
```
nikita-vanyasin, line 89: Rename Pod -> e.g. `%T` of spec?

pkg/util/compare/mode.go
```diff
@@ -0,0 +1,54 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Copyright holder is ArangoDB GmbH, Cologne, Germany
+//
+
+package compare
```
nikita-vanyasin, line 21: This `compare` package depends on deployment apis and is used only in `pkg/deployment`. Looks like `pkg/deployment/compare` is a better place for it?

nikita-vanyasin (commented):
LGTM, see suggestions

pkg/util/compare/mode.go
ajanikow: It will be used also by ML part, that's why it is extracted

nikita-vanyasin (approved):

PR #1494, opened 2023-11-17
[Feature] [ML] Handlers

ajanikow:


pkg/server/server.go
```diff
@@ -176,6 +177,18 @@ func NewServer(cli typedCore.CoreV1Interface, cfg Config, deps Dependencies) (*S
 		r.GET("/ready/storage", gin.WrapF(deps.Storage.Probe.ReadyHandler))
 		readyProbes = append(readyProbes, deps.Storage.Probe)
 	}
+	if deps.Backup.Enabled {
+		r.GET("/ready/backup", gin.WrapF(deps.Backup.Probe.ReadyHandler))
```
jwierzbo, line 181: shouldn't we add it to the liveness probes in the chart?

jwierzbo (approved):

cmd/cmd.go
```diff
@@ -119,6 +119,7 @@ var (
 		enableStorage               bool // Run local-storage operator
 		enableBackup                bool // Run backup operator
 		enableApps                  bool // Run apps operator
+		enableMl                    bool // Run ml operator
```
nikita-vanyasin, line 122:
```suggestion
		enableML                    bool // Run ml operator
```

nikita-vanyasin (approved):

pkg/server/server.go
ajanikow: It is by default part of /ready :) But now you can just ask for subhealth also

PR #1510, opened 2023-11-27
[Feature] [ML] Metadata Service Implementation

ajanikow:


pkg/apis/ml/v1alpha1/object.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	return &Object{
+		Name:      in.GetName(),
+		Namespace: util.NewType(in.GetNamespace()),
+		UID:       util.NewType(in.GetUID()),
+	}
+}
+
+type Object struct {
```
nikita-vanyasin, line 42: Why not in `shared/v1` package?

PR #1522, opened 2023-12-04
Switch to fsnotify

jwierzbo:
Previous lib was not compiling on MacOS

### DIFF
--- a/pkg/util/fwatch/watcher.go
+++ b/pkg/util/fwatch/watcher.go
@@ -108,6 +108,11 @@ func (fw *fileContentWatcher) Start(ctx context.Context) {
 					return
 				}
 
+				// File attributes were changed - skip it
+				if event.Op == fsnotify.Chmod {
+					continue
+				}
+
 				fw.log.Info("modified file: %s", event.Name)
 				fw.markAsChanged()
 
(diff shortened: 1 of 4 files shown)


pkg/util/fwatch/watcher.go
```diff
@@ -26,16 +26,11 @@ import (
...
 	"sync"
 
-	"k8s.io/utils/inotify"
+	"github.com/fsnotify/fsnotify"
 
 	"github.com/arangodb/kube-arangodb/pkg/logging"
 )
 
-const (
-	inotifyEventsModification = inotify.InCreate | inotify.InDelete | inotify.InDeleteSelf |
```
nikita-vanyasin, line 35: This is now not so optimal - there will be many events fired which does not mean that file content was actually changed.

But I guess this it is the price for cross-platform build, so LGTM.

nikita-vanyasin (approved):

pkg/util/fwatch/watcher.go
jwierzbo: This one should fix it:
```
if event.Op == fsnotify.Chmod {
	continue
}
```

nikita-vanyasin: @jwierzbo yeah, cool! Can you add that? feel free to merge afterwards

PR #1526, opened 2023-12-05
MLCronJob status update

jwierzbo:
https://github.com/arangodb/kube-arangodb-enterprise/pull/83


nikita-vanyasin (approved):

chart/kube-arangodb/templates/ml-operator/role.yaml
```diff
@@ -52,5 +52,9 @@ rules:
         - "secrets"
         - "serviceaccounts"
       verbs: ["*"]
+    - apiGroups: ["batch"]
```
ajanikow, line 55: Check few lines above

examples/ml/ml-cronjob.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: ml.arangodb.com/v1alpha1
```
ajanikow, line 1: Remove example, we will provide example in docs later

pkg/apis/ml/v1alpha1/extension_conditions.go
```diff
@@ -29,4 +29,7 @@ const (
 	ExtensionMetadataServiceValidCondition api.ConditionType = "MetadataServiceValid"
 	ExtensionServiceAccountReadyCondition  api.ConditionType = "ServiceAccountReady"
 	LicenseValidCondition                  api.ConditionType = "LicenseValid"
+	CronJobCreatedCondition                api.ConditionType = "CronJobCreated"
```
ajanikow, line 32: Remove conditions

ajanikow (changes requested):

examples/ml/ml-cronjob.yaml
jwierzbo: why? it is useful for quick tests...

ajanikow: We will generate it in examples and readme for all objects

ajanikow (approved):

PR #1537, opened 2023-12-12
Linter - sharedApi

jwierzbo:


nikita-vanyasin (approved):
LGTM.

Suggestion: we can introduce the linter rule for 
`github.com/arangodb/kube-arangodb/pkg/handlers/enterprise/ml/shared`
and
`github.com/arangodb/kube-arangodb/pkg/apis/shared` aliases too. Then we will never confuse those packages again.

i.e. `mlShared "github.com/arangodb/kube-arangodb/pkg/handlers/enterprise/ml/shared"`
`shared github.com/arangodb/kube-arangodb/pkg/apis/shared`

PR #1565, opened 2024-01-08
[Maintenance] Reduce binary size

ajanikow:


Makefile
```diff
@@ -62,6 +62,27 @@ GOASSETSBUILDER := $(GOBUILDDIR)/bin/go-assets-builder$(shell go env GOEXE)
...
+GOBUILDGCFLAGS :=
+
+# Go Strip Section
+GOBUILDSTRIP ?= 1
+ifeq ($(GOBUILDSTRIP),1)
+GOBUILDLDFLAGS += -w -s
+endif
+
+# Go Disable function inlining
+GOBUILDDISABLEFUNCTIONINLINING ?= 1
```
nikita-vanyasin, line 75: How much it affect the operator performance though? Wondering to see the benchmarking results - I think we have some benchmarking tests there.

PR #1573, opened 2024-01-11
[Bugfix] Don't abort plan in case of optional action timeout

nikita-vanyasin:


jwierzbo (approved):

pkg/deployment/reconcile/plan_executor.go
```diff
@@ -397,10 +397,14 @@ func (d *Reconciler) executeAction(ctx context.Context, planAction api.Action, a
 		d.context.CreateEvent(k8sutil.NewPlanAbortedEvent(d.context.GetAPIObject(), string(planAction.Type), planAction.MemberID, planAction.Group.AsRole()))
 		return false, true, false, false, nil
 	} else if isActionTimeout(timeout, planAction) {
+		if planAction.Type.Optional() {
+			log.Warn("Optional action not finished in time. Skipping")
+			return false, false, false, false, nil
```
ajanikow, line 402: Mark as ready, now Operator will just think action did not finish

ajanikow (approved):

ajanikow (changes requested):
Check comments

pkg/deployment/reconcile/plan_executor.go
nikita-vanyasin: Done

ajanikow (approved):

PR #1634, opened 2024-03-26
[Feature] Scheduler BatchJob Integration Definition

ajanikow:


jwierzbo (approved):

neunhoef (commented):
This looks OK at first glance. However, I am missing the place where more details about the job at hand (what algorithm, what container limitations, what input data, what place to write the output to) can be specified. The message "BatchJobSpec" has some meta information but no space to actually specify more details.
Maybe I am missing context here? Do we have more documentation or design docs about the whole intended system?

geenen124 (approved):

PR #1641, opened 2024-04-08
[Feature] Parametrize Scheduling Graceful Duration

ajanikow:


jwierzbo (approved):

pkg/util/globals/global.go
```diff
@@ -29,6 +29,7 @@ const (
 	DefaultArangoDCheckTimeout                = time.Second * 2
 	DefaultReconciliationTimeout              = time.Minute
 	DefaultForcePodDeletionGracePeriodTimeout = 15 * time.Minute
+	DefaultPodSchedulingGracePeriod           = 15 * time.Second
```
jwierzbo, line 32: IMO it could be bit longer (60s)

PR #1690, opened 2024-07-29
[Feature] ArangoRoute RC

ajanikow:


chart/kube-arangodb/values.yaml
```diff
@@ -34,6 +34,7 @@ operator:
     k8sToK8sClusterSync: false
     ml: false
     analytics: false
+    networking: true
```
jwierzbo, line 37: why this is enabled by default?

jwierzbo (approved):

chart/kube-arangodb/values.yaml
ajanikow: As designed - will be required for core functionality

PR #1694, opened 2024-08-14
OAS-9904 Introduce ArangoDeployment Gateway Group

jwierzbo:


pkg/deployment/resources/pod_creator.go
```diff
@@ -286,6 +286,17 @@ func createArangoSyncArgs(apiObject meta.Object, spec api.DeploymentSpec, group
 	return args
 }
 
+func createArangoGatewayArgs(groupSpec api.ServerGroupSpec) []string {
+	args := []string{
+		fmt.Sprintf("-c %s", GatewayConfigFilePath),
```
ajanikow, line 291: Single param?

pkg/deployment/resources/pod_creator.go
```diff
@@ -379,6 +390,21 @@ func (r *Resources) RenderPodForMember(ctx context.Context, acs sutil.ACS, spec
 			memberStatus: m,
 			cachedStatus: cache,
 		}
+	case api.ServerGroupTypeGateway:
+		imageInfo.Image = r.context.GetOperatorImage()
```
ajanikow, line 394: Is it possible to get .spec.gateway.image ? IF no, please create ticket for that

pkg/deployment/resources/pod_creator.go
jwierzbo: ?

pkg/deployment/resources/pod_creator.go
jwierzbo: Ticket created

pkg/deployment/resources/pod_creator.go
ajanikow: Please, split it to:
- full param name (instead of -c)
- 2 params instead of 1 (so list with 2 elements)

ajanikow: []string{"--config...", GatewayConfigFilePath}

ajanikow (approved):

PR #1768, opened 2024-11-15
Update doc to have a arango-root-pwd that works

GregTwinsight:


ajanikow:
Hello!

Can you please sign the NDA?

If not, we will fix it with a doc update for the next Release.

BR.
Adam.

PR #1903, opened 2025-06-02
Fix debug doc

jwierzbo:


ajanikow (changes requested):
Flag has been added as designed to match the namespace of dump - so in best case it should be present in both places

ajanikow (approved):

PR #2099, opened 2026-05-20
[Feature] [RBAC] Design adjustments

ajanikow:


docs/platform/rbac/faq.md
```diff
@@ -0,0 +1,175 @@
...
+parent: RBAC
+grand_parent: ArangoDBPlatform
+nav_order: 6
+---
+
+# RBAC FAQ
+
+## General
+
+### How do I check if RBAC is enabled?
```
cmyk47, line 13: can we get an endpoint to check this? e.g. `GET /_management/permissions/status` -> status: `200`, body `{ "enabled": true }`

cmyk47: _context_: I need to show or hide the RBAC section in the UI (sidebar, menus, etc), so I need a way to check whether it's active from something the UI can actually reach (like an endpoint, not from Kubernetes)

docs/platform/rbac/identity.md
```diff
@@ -0,0 +1,193 @@
+---
+layout: page
+title: Identity and Permissions
+parent: RBAC
+grand_parent: ArangoDBPlatform
+nav_order: 5
+---
+
+# Identity and Permission Checks
```
cmyk47, line 9: These endpoints (`identity`, `login`, `logout`) look identical to the ones the UI already calls today at `/_identity`, `/_login`, `/_logout` seems to be same request bodies and response shapes.

Which set should the UI use going forward: the short `_` paths, or `/_integration/authn/v1/*`?

docs/platform/rbac/identity.md
```diff
@@ -0,0 +1,193 @@
...
+Response:
+```json
+{
+  "user": "alice"
+}
+```
+
+If no user is specified in the token, `root` is returned.
+
+### POST `/_integration/authn/v1/validate`
```
cmyk47, line 64: Are the other endpoints in this doc `validate`, `createToken`, `permissions`/`validate`, `authorization/v1/evaluate*`, intended to be called from the UI, or are they for backend / service-to-service use?

I'm trying to figure out which we need to wire into the UI for the initial release. My current understanding is that since the first version of RBAC in the UI only needs to display permission-denied messages, we won't need any of these, but I'd like to confirm that before moving on.

docs/platform/rbac/user_bindings.md
```diff
@@ -0,0 +1,135 @@
...
+
+User role bindings assign roles to specific users with per-user scope policies.
+This allows the same role to grant different permissions depending on who holds
+it.
+
+## API Endpoints
+
+All endpoints require authentication and the appropriate RBAC permission.
+
+### List bindings for a user
```
cmyk47, line 19: Is there a list-all-bindings endpoint, or only the per-user `GET /user/{user}/role`? Without one, the Users tab would have to do an N+1 fan-out (one request per user) to render the list, a `GET /user/role` or `?users=user1,user2,etc` variant would help

ajanikow: We can't provide a way to list all bindings - they are per role/user - in order to make it secure and limited, we need to split API calls.

cmyk47: could you help me understand the threat model here? If a caller is authorized to call `GET /user/{user}/role` for every user, they can reconstruct the full binding list anyway, just with N requests instead of one. The authorization check is the same either way, right?

if theres a different concern driving this, happy to work around it, just want to understand it first

docs/platform/rbac/faq.md
cmyk47: with the update cb7ea6c115c63f5efb4a2899840cc90125e5543f now is pretty clear. Thanks!

design/rbac/api.md
```diff
@@ -0,0 +1,142 @@
...
+Response: `AuthorizationAPIValidateResponse { message, effect }`
+
+Implementation: `pkg/sidecar/services/authorization/impl_validate.go`
+- Extracts identity from gRPC context via `authenticator.GetIdentity(ctx)`
+- Calls `EvaluatePermission()` on the identity with the plugin chain
+- No RBAC permission required (self-check only)
+
+### Policy Management
+
+| Method | Path | RBAC Action |
```
mchacki: Please add one column for "Purpose" / when to use this Action.

it is kind of obvious here if you know how CRUD works, but I prefer to have this small description.
It may also help LLMs to pick the right option easier.

design/rbac/api.md
```diff
@@ -0,0 +1,142 @@
...
+| `GET` | `/policy/{name}` | `rbac:GetPolicy` |
+| `POST` | `/policy/{name}` | `rbac:CreatePolicy` |
+| `PUT` | `/policy/{name}` | `rbac:UpdatePolicy` |
+| `DELETE` | `/policy/{name}` | `rbac:DeletePolicy` |
+
+Implementation: `pkg/sidecar/services/authorization/impl_api_policy.go`
+
+### Role Management
+
+| Method | Path | RBAC Action |
```
mchacki: See above comment

design/rbac/api.md
```diff
@@ -0,0 +1,142 @@
...
+| `GET` | `/role/{name}` | `rbac:GetRole` |
+| `POST` | `/role/{name}` | `rbac:CreateRole` |
+| `PUT` | `/role/{name}` | `rbac:UpdateRole` |
+| `DELETE` | `/role/{name}` | `rbac:DeleteRole` |
+
+Implementation: `pkg/sidecar/services/authorization/impl_api_roles.go`
+
+### User Role Binding Management
+
+| Method | Path | RBAC Action |
```
mchacki: See above comment

design/rbac/api.md
mchacki: Also please add the response types. What do I get back when calling this API

design/rbac/api.md
```diff
@@ -0,0 +1,142 @@
+# RBAC API
+
+The RBAC system exposes two API surfaces: the **management API** (for managing
+policies, roles, and bindings) and the **integration API** (for authentication
+and permission evaluation).
+
+## Management API (`/_management/permissions/`)
```
mchacki: In this section the path is ambiguous.

Is this a prefix for all below routes?
So the first path in this document would be:
`/_management/permissions/permissions/validate`

the double use of permissions directly after one another is confusing to read.
I do See two options here:
1) Make path below explicit including prefix (safe)
2) Swap Validation with one of the other topics, then the brain pattern matching does not kick in. ;)

mchacki: Is there no version in this API?
the other api has a version number

design/rbac/api.md
```diff
@@ -0,0 +1,142 @@
...
+Bindings are keyed as `urb:<user>:<role>`. The list endpoint scans by prefix
+`urb:<user>:` to find all bindings for a user.
+
+## Integration API
+
+### Authentication (`/_integration/authn/v1/`)
+
+Exposed by the authentication integration service.
+
+| Method | Path | Description |
```
mchacki, line 89: Aha this description is good, and what I am missing on above.
Please add response here as well.

design/rbac/api.md
```diff
@@ -0,0 +1,142 @@
...
+|---|---|---|
+| `GET` | `/identity` | Returns the current user's username |
+| `POST` | `/validate` | Validates a token, returns user, roles, and lifetime |
+| `POST` | `/login` | Authenticates with username/password, returns JWT |
+| `GET` | `/logout` | Clears cookies, redirects |
```
mchacki: redirects where?

design/rbac/api.md
mchacki: All descriptions in this document also miss input bodies.

design/rbac/api.md
mchacki: (I agree this is a bit of nit picking, you get it when reading the full doc) So I am good with a "won't fix" here

design/rbac/architecture.md
```diff
@@ -0,0 +1,93 @@
...
+```
+
+## Envoy ext_authz
+
+HTTP requests hit the Envoy gateway, which sends a `CheckRequest` to the
+ext_authz gRPC service. A handler chain processes the request through
+authentication stages: bearer token, cookie, OpenID Connect, session handling,
+and pass-mode configuration. Each route can override auth behavior via
+`ExtAuthzPerRoute` context extensions.
+
```
mchacki, line 48: Please add one or two lines in which stage the user is now in.
We know this user is valid?
We know this users Roles?
We do not know if the user can access what she is asking for?

design/rbac/architecture.md
```diff
@@ -0,0 +1,93 @@
...
+The plugin chain determines how authorization decisions are made:
+
+1. **Superuser check** - Nil user (service accounts) always granted
+2. **Plugin dispatch** - Routes to the configured plugin mode:
+   - `Central` - Strict policy evaluation
+   - `CentralPermissive` - Wraps Central, converts errors to Allow
+   - `Always` / `Never` - Bypass modes
+3. **Policy evaluation** - Collects and evaluates all applicable policies
+
```
mchacki, line 87: What does this do?
Does this modify gRPC context?
Does it already reject users?

design/rbac/architecture.md
```diff
@@ -0,0 +1,93 @@
...
+Key files:
+- `integrations/authorization/v1/shared/authz.go` - `Plugin` interface
+- `integrations/authorization/v1/shared/superuser.go` - Superuser bypass
+- `integrations/authorization/v1/shared/permissive.go` - Error-tolerant wrapper
+- `integrations/authorization/v1/configuration.go` - Configuration types
+
+## Feature Flags
+
+RBAC is controlled by two feature flags that determine the authorization mode
```
mchacki, line 96: Please add a short description on where to set those Feature Flags.

e.g. as injected in the cli via `--<option>`

design/rbac/api.md
```diff
@@ -0,0 +1,142 @@
...
+|---|---|---|
+| `GET` | `/user/{user}/role` | `rbac:ListUserRoleBinding` |
+| `POST` | `/user/{user}/role/{role}` | `rbac:AssignUserRole` |
+| `DELETE` | `/user/{user}/role/{role}` | `rbac:RemoveUserRole` |
+| `PUT` | `/user/{user}/role/{role}` | `rbac:ReplaceUserRoleScope` |
+
+Implementation: `pkg/sidecar/services/authorization/impl_api_user_role_bindings.go`
+
+Bindings are keyed as `urb:<user>:<role>`. The list endpoint scans by prefix
+`urb:<user>:` to find all bindings for a user.
```
mchacki, line 81: Can you add a section about auditing?
Can I activate auditing on Permission changes?
If so, how can I access the Audit log?

design/rbac/evaluation.md
```diff
@@ -0,0 +1,79 @@
...
+
+1. `getUserPolicies(user, roles...)` iterates all stored roles, matches by
+   role name, parses named policies, and includes scope policies.
+
+Key file:
+- `pkg/sidecar/services/authorization/client.go`
+
+## Configuration Modes
+
+The authorization plugin mode is set per deployment:
```
mchacki, line 103: Where do I set this configuration?
Can I change it at runtime?
Should I ever change this?

design/rbac/evaluation.md
```diff
@@ -0,0 +1,79 @@
...
+| `Central` | Strict enforcement. Deny on evaluation failure. |
+| `CentralPermissive` | Central with fallback to Allow on error. Logs the error. |
+| `Always` | Always returns Allow. Used for development/testing. |
+| `Never` | Always returns Deny. Locks out all access. |
+
+### Superuser Bypass
+
+Requests with a nil user identity (no username in the JWT, typically internal
+service accounts) bypass policy evaluation entirely and are always granted.
+Named users always go through the full evaluation.
```
mchacki: Please write a short NOTE here:
That `nil` users cannot be created from the outside and are blocked by authentication.

design/rbac/storage.md
```diff
@@ -0,0 +1,74 @@
+# RBAC Storage
+
+## ArangoDB Collections
+
+All authorization objects are stored in ArangoDB system collections derived
+from the `_users` collection properties:
```
mchacki, line 6: This includes the following:

* They are only present in the `_system` database and hence managed globally, and not per database.
* They use a single shard.

design/rbac/storage.md
```diff
@@ -0,0 +1,74 @@
...
+- Provides CRUD operations (`Create`, `Update`, `Delete`, `Item`)
+- Supports change polling via `Pool(start)` for streaming updates
+- Runs a background goroutine that periodically refreshes from the database
+
+The pooler timeout is 10 seconds by default (`DefaultPoolerTimeout`).
+
+Key file:
+- `pkg/sidecar/services/authorization/pool/pool.go`
+
+## Client-Side Sync
```
mchacki, line 35: This section seems to contains some guarantees.

What is the "delay" interval?

design/rbac/storage.md
```diff
@@ -0,0 +1,74 @@
...
+
+1. **Initial load** - Calls `GetPolicy()` / `GetRole()` stream RPCs to fetch
+   all current items.
+2. **Continuous polling** - Calls `PoolPolicyChanges()` / `PoolRoleChanges()`
+   with the last known index and a 15-second timeout.
+3. **Cache rebuild** - On each update, the `internalCache` is rebuilt with
+   parsed policies and role-to-policy mappings.
+
+The client considers itself unhealthy if either the policies or roles haven't
+been updated within the last minute.
```
mchacki: What is the consequence here?
Will I get outdated authorization?
Will I get logged out of the system by "deny everything" ?

design/rbac/token_integration.md
```diff
@@ -0,0 +1,86 @@
...
+(named `managed:operator:<uid>`). This is a convenience for simple use cases
+where a single token needs self-contained permissions.
+
+For manual, fine-grained RBAC management, use the dedicated binding CRDs
+instead:
+
+- **ArangoPermissionPolicy** - Define reusable policies
+- **ArangoPermissionRole** - Define roles with inline scope
+- **ArangoPermissionPolicyRoleBinding** - Attach policies to roles
+- **ArangoPermissionRoleUserBinding** - Assign roles to users with per-user scope
```
mchacki: Okay here I get confused.
This CRDs look like the static configuration.
And then we have some data stored in Arango itself.
Where are they overlapping?

I did understand it the following way:
We use CRDs to define which Policies exist, and to which role they belong.
But the whole "User X has Role Y" is stored in the database?
Is that an incorrect understanding?

If it is correct what is `ArangoPermissionRoleUserBinding` for?

design/rbac/token_integration.md
```diff
@@ -0,0 +1,86 @@
...
+- Username
+- Signing secret hash
+- TTL
+- Roles list (including managed role name when `status.role` is set)
+
+When `HandleArangoDBRole` sets `status.role` (after creating the role), the
+hash changes on the next reconcile cycle. `HandleArangoSecret` detects the
+mismatch and regenerates the JWT with the managed role included.
+
+This means there is a **one-cycle delay** between role creation and the JWT
```
mchacki: This is a very important guarantee or impact.

design/rbac/token_integration.md
```diff
@@ -0,0 +1,86 @@
...
+3. **HandleArangoDBPolicy** - Renders `spec.policy` into a proto `Policy`
+   object and creates/updates it in the authorization sidecar as
+   `managed:operator:<uid>`.
+
+4. **HandleArangoDBRole** - Creates a `Role` referencing the named policy with
+   the rendered policy as its scope. Stored as `managed:operator:<uid>`.
+
+## JWT Regeneration
+
+The JWT includes a hash computed from:
```
mchacki, line 129: Just to be clear:
The JWT does only contain the Role Name, but not the Scope?

e.g. i am a DBAdmin only for database foo.
Then my Role will be DBAdmin?
Or will it be some internal pattern like `DBAdmin_foo` ?

design/rbac/architecture.md
```diff
@@ -0,0 +1,93 @@
+# RBAC Architecture
```
mchacki, line 1: I do not think this comment fits into this File, but it is a close it could get.

I would like to ask for the following two documents:
1) a Summary document for "clients". Containing the overview of the following:
a) When I modify Roles of a user, what is the timeframe we guarantee this takes effect everywhere in the platform?
b) When I modify my own Roles, will I get immediate update? Will it take some timeframe? Do I need to logout and login?
c) What happens in case of the RBAC system is not reachable? (e.g. there is a network split and system is in degraded state, or the connection of one pod to the RBAC is broken)

2) A "How To" document for ArangoDB implementors of services. What do I as a developer need to do if I want to make my Service RBAC ready? (I am good if this is a confluence page or otherwise internal, and we just add a link there)

docs/platform/rbac/enabling.md
```diff
@@ -0,0 +1,87 @@
...
+| central-services | rbac-enforced | Mode | Behavior |
+|---|---|---|---|
+| Disabled | - | `always` | All requests allowed (no authorization) |
+| Enabled | Disabled | `central-permissive` | Policies evaluated but denials are logged, not enforced |
+| Enabled | Enabled | `central` | Full enforcement, denied requests are blocked |
+
+### Recommended Rollout
+
+1. **Start with permissive mode** - Enable `central-services` without
+   `rbac-enforced`. This evaluates policies and logs denials without blocking
```
mchacki, line 62: Those logs are where?
How can I access them?

docs/platform/rbac/policies.md
```diff
@@ -0,0 +1,306 @@
...
+The recommended flow for setting up RBAC via CRDs is:
+
+1. Create a policy (`ArangoPermissionPolicy`)
+2. Create a role (`ArangoPermissionRole`)
+3. Bind the policy to the role (`ArangoPermissionPolicyRoleBinding`)
+4. Bind the role to a user (`ArangoPermissionRoleUserBinding`)
+
+## Managing via API
+
+### Create a policy
```
mchacki, line 134: Are those publicly available APIs?

AFAIK the first version should not allow this to be set by a client.

mchacki (commented):
I have some open Questions on the design

docs/platform/rbac/enabling.md
ajanikow: Just logs in the Pod

design/rbac/token_integration.md
ajanikow: Removed, as we will move different way

mchacki (approved):
LGTM

PR #2105, opened 2026-06-08
[Design] [Platform] OpenTelemetry integration

ajanikow:


design/platform/monitoring/otel.md
```diff
@@ -0,0 +1,194 @@
...
+        sidecar:
+          env:
+            - name: OTEL_SERVICE_NAME
+              value: "arangodb-sidecar"
+```
+
+## How It Works
+
+1. The `ArangoProfile` is matched to pods via label selectors
+2. The profile template's `container.all.env` entries are merged into every
```
juergenkaiser-dev, line 166: What happens if the container already has those env variables defined?

ajanikow: We override envs in the Pods using profiles, but envs should not be present

jwierzbo (approved):

PR #2140, opened 2026-08-03
Allow WebSocket upgrades over HTTP/2 on the gateway

ajanikow:
## What

Enables WebSocket upgrades over HTTP/2 on the ArangoDB gateway (Envoy) downstream listener by turning on **RFC 8441 Extended CONNECT**:

```go
func listenerHttp2ProtocolOptions() *pbEnvoyCoreV3.Http2ProtocolOptions {
    return &pbEnvoyCoreV3.Http2ProtocolOptions{AllowConnect: true}
}
// ...set as HttpConnectionManager.Http2ProtocolOptions
```

This is consistent with the gateway already defaulting to HTTP/2 upstream — without `allowConnect`, WebSocket `CONNECT`-based upgrades cannot be tunneled over an HTTP/2 connection.

## Test

`Test_GatewayConfig` now asserts, via `requireListenerHTTP2AllowConnect`, that the rendered HttpConnectionManager has `Http2ProtocolOptions.AllowConnect == true` in the relevant subtests.

## Note

This covers the downstream handshake. End-to-end WebSocket proxying additionally requires a `websocket` upgrade config on the route/HCM and an upstream that supports it; if those are not already present a follow-up may be needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Change Log
 
 ## [master](https://github.com/arangodb/kube-arangodb/tree/master) (N/A)
+- (Feature) (Gateway) Allow WebSocket upgrades over HTTP/2 by enabling RFC 8441 Extended CONNECT (allowConnect) on the gateway downstream listener
 - (Bugfix) Strip the pre-release/build suffix (e.g. `-devel`) from the detected ArangoDB version so version and feature gates compare numerically instead of mis-ordering nightly builds below release versions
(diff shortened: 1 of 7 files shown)


jwierzbo (approved):

ajanikow:
E2E test coverage in the test repo: arangodb/kube-arangodb-test#436 (three cases — no-upgrade, feature-disabled, feature-enabled — over HTTP/1.1 and HTTP/2).

PR #2158, opened 2026-08-19
Reload gateway TLS certificates in place via Envoy SDS, validate gateway serving certs, and clean up the member keyfile on removal

ajanikow:
Three related fixes to the gateway TLS lifecycle.

## 1. Reload TLS certificates in place via Envoy SDS
Previously the listener referenced the internal serving keyfile and every SNI keyfile as inline `DataSource_Filename` paths. The gateway config checksum is computed over the rendered YAML, which references certs only by path, so **rotating a secret's contents changed neither the LDS listener nor the config checksum** — a static gateway never restarted and a dynamic gateway never got an LDS change, so a rotated cert was silently missed.

Now each certificate (internal + one per SNI) is delivered through filesystem **SDS**: the listener uses `tls_certificate_sds_secret_configs` → `PathConfigSource{ Path, WatchedDirectory }`, where `WatchedDirectory` is the mounted secret directory (`/secrets/tls`, `/secrets/sni/<secret>`). Envoy hot-reloads the certificate in place on rotation — no listener churn, no pod restart. The SDS `Secret` definitions are rendered into a new `<depl>-gateway-sds` ConfigMap, mounted at `/etc/gateway/sds/` when TLS is enabled.

## 2. Validate gateway serving certificates like arangod members
Gateways are excluded from `AllArangoDServerGroups`, so their serving certificate was created but **never validated or renewed** the way coordinator/arangod certs are. The keyfile-renewal check (endpoint verification, expiry-margin, alt-name match) is now also run for the gateway group; on failure it removes the keyfile and restarts the gateway (recreate path), which regenerates the keyfile with the correct alt names. `keyfileRenewalRequired`'s group switches learned the gateway case (alt names via `GetServerAltNames`; the arangod in-place propagation check is skipped — Envoy reloads via SDS from fix #1).

## 3. Clean up the member TLS keyfile secret on gateway removal
Gateway Pods now carry a `gateway.database.arangodb.com/tls-keyfile` finalizer (when TLS is enabled). The keyfile secret is owner-referenced to the ArangoMember and is garbage-collected on member removal; the finalizer adds a prompt best-effort cleanup on a genuine teardown (member marked to remove) and always releases once the pod is terminating, so it never blocks a rolling restart.

## Tests
- New `Test_GatewaySDS`: asserts the listener delivers certs via SDS (not inline) with the correct watched directory, and that `RenderSDS` emits one definition file per certificate.
- Existing gateway render (determinism), resources and reconcile TLS suites pass; full `go build ./...`, `go vet`, gofmt/goimports clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Change Log
 
 ## [master](https://github.com/arangodb/kube-arangodb/tree/master) (N/A)
+- (Feature) Deliver the gateway's TLS certificates (internal and SNI) to Envoy via filesystem SDS with a watched directory, so a rotated certificate is reloaded in place without restarting the gateway
+- (Feature) Validate gateway serving certificates (endpoint verification, expiry margin and alt-name match) like arangod members and trigger keyfile renewal + restart when required
 - (Maintenance) Add a finalizer to gateway Pods that cleans up the member TLS keyfile secret when the gateway member is removed
 - (Maintenance) Lint all Helm charts in CI via `make helm-lint`; fix the `platform-storage` passwords template document separator and add the missing `apiVersion` to the `kube-arangodb-crd` chart
 - (Documentation) Add a README for the `platform-storage` chart and reference it from `docs/helm.md`, the MinIO storage-integration docs and the main README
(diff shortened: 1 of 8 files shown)


ajanikow:
Split into focused PRs: finalizer → #2159 (merged), SDS reload → #2161, cert validation → #2162. Closing this combined PR as superseded.

